PR #1133, opened 2013-02-15
Select - optgroup support

lordbron:
In case you like them both and want an easy way to pull 'em both in.


thibaultzanini:
@lordbron Can you please update this PR with the latest version of Montage.

PR #1204, opened 2013-06-18
Use more sensible text coercion operator in TextField

kriskowal:
The TextField would coerce objects, null, and undefined with the
JavaScript convention. This is not a good user experience in general, so
I've reused the string coercion operator from FRB, which translates
these all to an empty string.


rayshan:
reopen per comments

hthetiot:
Fixed by 7eed482e2e6b4afe68fe6284bc12afc2da1e909a

PR #1240, opened 2013-07-31
Fix Overlay positioning problems with anchors

romancortes:


rayshan:
@romancortes Hi Javier can you take a look and decide whether this should be worked on or closed?

PR #1267, opened 2013-08-26
Build montage.js bootstrapping using a build script

kriskowal:
:warning: This depends on changes to `mr` in the corresponding `build` branch, which has not landed at time of writing.  Changes to `mop` in the corresponding `build` branch are also required for compatibility.  As such, the `build` branches of all of these repositories must be coordinated and land in a backward-incompatible release.

`mr` now includes a `mrs` script for building some stand-alone scripts based on Mr’s dependency analysis if the built modules only need the ability to require and export modules in a static working set.  The resulting script has very little overhead for linking the modules.  Lazy loading and compiler extensions are not supported, but would be possible with future work.

`mrs` is suitable for building the new Mr `boot.js` and the Montage boot script, `montage.js`.  This change set introduces the command `npm run build` that updates `montage.js` and also produces `boot/preload-boilerplate.js`.

This introduces the need to rebuild `montage.js` whenever modules in the `boot` directory, `mr`, or any of their transitive dependencies are modified.

This obviates the need to perform script injection in `montage.js`, reducing code size and complexity in both development and production.

This obviates the need to have the `packages` checked into Git and the sub-repository update machinations that entails.  These dependencies can be installed normally, with NPM.  The relevant modules are incorporated into `montage.js` by the build script.  Since `mr` and its dependencies now exist in `node_modules`, this obviates the need for Montage's `package.json` to use Mr-specific `mappings` or `overlays`.

This reduces the scope of `montage.js` to only need to support the browser script loading use case.  It no longer needs to also serve as a Node.js module.  It no longer needs to also serve as the boilerplate for bundling.  These cases are now served in simpler and separate ways.

The parts of `montage.js` that were needed to bootstrap in Node.js are now in CommonJS modules in the `boot` directory.  These are embeded in `montage.js` by the build script, but can also be used as Node.js modules without any wrappers or boilerplate.

The parts of `montage.js` that are needed to bootstrap in bundles built by `Mop` are now in `boot/preload-boilerplate.js`, which like `montage.js` is built using `mrs`.

This obviates the need to have a hard-coded copy of `mr/mini-url.js`.  Mr and Montage can now share this CommonJS module instead of embedding it in their bootstrapping, which will make that function easier to maintain going forward.  This also obviates the need for the `mini-url` dependency to be handled specially at run-time.  Regardless of the platform, modules may depend on `url`.  When used in a browser, the module system will link to `mini-url.js`.  In Node.js, where `mini-url.js` would not work at all, it falls back to the `url` module provided by the platform.

When Mop prepares a bundle, it is now necessary for it to provide a `data-location` property instead of `data-montage-location` or `data-mr-location` since the code that handles this detection is now unified in `mr/boot/script-params.js`.


rayshan:
Closing, will revisit when we decide to have a build step.

hthetiot:
We going another direction.

PR #1285, opened 2013-09-04
Add type inference engine

kriskowal:
This is a type inferrence engine that takes an FRB expression and the types of objects in scope and makes a best-effort attempt at determining the type of the expression.

This is not in its final form, but it certainly is needed to do longitudinal development on data-driven components.


rayshan:
Hi @kriskowal, can you please comment on how close this is to be integrated?

PR #1393, opened 2014-01-08
Blueprint omnibus

kriskowal:
This consolidates changes form #1215 and #1309. Posted for another round of review, mostly so @spearway can verify that these are his intended changes. This merge was performed by @kriskowal and @Stuk sharing a screen.

### DIFF
--- a/core/core.js
+++ b/core/core.js
@@ -1190,7 +1190,8 @@ exports._blueprintDescriptor = {
                         // requested does not exist. If any parents do not
                         // exist then the error should still be thrown.
                         if (error.message.indexOf("Can't XHR") !== -1) {
-                            return Blueprint.createDefaultBlueprintForObject(self).then(function (blueprint) {
+                            return Blueprint.createDefaultBlueprintForObject(self)
+                            .then(function (blueprint) {
                                 return blueprint;
                             });
                         } else {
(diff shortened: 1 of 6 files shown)


cdebost:
This is going to take some rebasing. Will take a look after we merge #1827. Could be a the beginning of a fix for https://github.com/montagejs/montage/issues/1832.

cdebost:
There are a few different things being changed in this PR:
1. Remove circular references (property & event descriptors pointing to their owner object descriptor).
2. Replace ObjectDescriptor's property & event containers (currently arrays) with Set to do uniquing for us.
3. Allow us to configure an ObjectDescriptor's constructors for creating object properties, property descriptors, etc.
4. Add versioning to descriptors. The intention is to be able to handle different serialization formats and deserialize differently depending on the version, and make sure older versions of montage don't try to deserialize a serialization they can't understand. I don't see why this is necessary for these changes considering all should be backwards compatible, and we don't have much incentive to ensure older versions of Montage can deserialize serializations made with newer versions.
5. Lots of non-consequential renaming and reformatting.

I think I'm going to cherry pick some of these changes (1 and 2) for application on master and leave the rest unless we can find a use case for which we would want them.

PR #1432, opened 2014-02-26
Add Element Component

kriskowal:
That is all.

### DIFF
--- a/ui/element.reel/element.js
+++ b/ui/element.reel/element.js
@@ -1,0 +1,7 @@
+
+var Component = require("../component").Component;
+
+exports.Element = Component.specialize({
+    hasTemplate: {value: false}
+});
+


cdebost:
FWIW, this adds a new component just so we can write:

```
{
    "someElement": {
        "prototype": "montage/ui/element",
        "properties": {
            "element": {"#": "someElement"}
        },
        "bindings": {
            "classList.has('foo')": {"<-": "@owner.foo"}
        }
    }
}
```

instead of:

```
{
    "someElement": {
        "prototype": "montage/ui/component",
        "properties": {
            "element": {"#": "someElement"},
            "hasTemplate": false
        },
        "bindings": {
            "classList.has('foo')": {"<-": "@owner.foo"}
        }
    }
}
```

As this stands I'm not sold on the added value here.

marchant:
If/when we complete the work that Thibault started to turn a Component in a differed graphic context, we should be using "value": to refer to the DOM element, and then use values:{} . We shouldn't need to create an additional element/component like this. Thibault, can you check how much more work was needed to complete the spec?

thibaultzanini:
@benoitmarchant I can confirm you that this spec doesn't work at the moment with value. Let's create a ticket

```
"someElement": {
                "value":  {"#": "someElement"},
                "bindings": {
                    "classList.has('foo')": {"<-": "@owner.foo"}
                }
            }
```

thibaultzanini:
Closing it, issue will be handle by https://github.com/montagejs/montage/pull/1845

PR #1481, opened 2014-06-11
Prevent preventDefault from always being executed.

3on:
I can't find any tests for the `KeyManager`.


thibaultzanini:
Already been taking care of here:
https://github.com/montagejs/montage/commit/b06ac8347164eed9881e5f6bc420129abf2c7bb1

PR #1509, opened 2015-01-02
Fix an issue with the substitution when the switchValue is equal to 0.

thibaultzanini:
Make sure that number is not implicitly converted to a Boolean value after the content did change. This behaviour was blocking the fact to save the content of the “SwitchComponent” related to this value.


rayshan:
Per discussion w/ @thibaultzanini:

- Add an `else` block to  `console.warn` user
- Add test for `it("should only switch when switchValue is a String or a Number")`

thibaultzanini:
Cannot reproduce. closing...

PR #1510, opened 2015-01-13
Fix for a bug that prevented a repetition clear selection to work.

marchant:


thibaultzanini:
@marchant It looks like to fix some specs but brings some new.
https://github.com/montagejs/montage/blob/master/test/run.js#L46

rayshan:
Random question for style guide - why `EMPTY_ARRAY`, not `emptyArray`?

thibaultzanini:
I assume it's because we want to distinguish constant variables. @marchant Can you confirm it?

rayshan:
I think you're right, @marchant mentioned it in passing. Thanks @thibaultzanini

marchant:
All range-controller tests now pass, merging

PR #1511, opened 2015-01-19
Fix unit tests

pchaussalet:


rayshan:
Hi @pchaussalet, can we have some description regarding these 2 commits? Not clear what the original issues were based on the commit message.

- Fix range-controller to respect constraint avoidsEmptySelection
- Tree-controller now support being initialized with a whole tree as content

rayshan:
@pchaussalet I'm cherrypicking your test fixes into another PR

hthetiot:
Too touchy next milestone

PR #1512, opened 2015-01-23
Fix event binding bug in swipe-composer

kevinluosl:
Fix in swipe that should use this._element instead document to bind event to target element

### DIFF
--- a/composer/swipe-composer.js
+++ b/composer/swipe-composer.js
@@ -11,17 +11,17 @@ var Montage = require("../core/core").Montage,
  * @classdesc `Composer` for detecting swipe gestures.
  * @extends Composer
  */
-exports.SwipeComposer = Composer.specialize( /** @lends SwipeComposer# */ {
+exports.SwipeComposer = Composer.specialize(/** @lends SwipeComposer# */ {
 
     load: {
         value: function() {
-            document.addEventListener("touchstart", this, true);
+            this._element.addEventListener("touchstart", this, true);
         }
     },
 
     unload: {
         value: function() {
-            document.removeEventListener("touchstart", this, true);
+            this._element.removeEventListener("touchstart", this, true);
         }
     },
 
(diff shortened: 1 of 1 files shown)


rayshan:
Just curious, is there a difference b/t `this._element` and `this.element`?

kevinluosl:
From the codes there is no different. When I was doing this modify, I saw other composer using this._element so I followed.

rayshan:
My experience is that they work the same as well. I'll ask around.

rayshan:
I discussed w/ @thibaultzanini, summary:

- Don't get or set `_privateProperty` directly due to possible custom getter / setters
- Can call methods on `_privateProperty` directly for better performance, e.g. `this._element.addEventListener`

I'll add this to docs at some point.

composer/swipe-composer.js
```diff
@@ -193,7 +199,7 @@ exports.SwipeComposer = Composer.specialize( /** @lends SwipeComposer# */ {
                 return;
             }
 
-            document.removeEventListener("touchmove", this, true);
+            this._removeEventListeners();
```
rayshan: Why do this right here, not the end of `captureTouchend` block?

rayshan:
Hi @kevinluosl, can we have some background on why events should be on `this._element` instead `document`? If we're using `this._element`, should we still use capture instead of the default bubble?

composer/swipe-composer.js
```diff
@@ -78,9 +78,9 @@ exports.SwipeComposer = Composer.specialize( /** @lends SwipeComposer# */ {
             this._startX = touch.clientX;
             this._startY = touch.clientY;
             this._startTimestamp = event.timeStamp;
-            document.addEventListener("touchmove", this, true);
-            document.addEventListener("touchend", this, true);
-            document.addEventListener("touchcancel", this, true);
+            this._element.addEventListener("touchmove", this, true);
+            this._element.addEventListener("touchend", this, true);
+            this._element.addEventListener("touchcancel", this, true);
```
thibaultzanini: agreed for listening on the element swipe composer for unload and load but not for touchmove, touchend, and touchcancel

composer/swipe-composer.js
thibaultzanini: I'm not sure If it's a big deal or not, but yeah in order to keep the same logic in our framework (we chosen that way for the translate composer for exemple)

thibaultzanini:
@kevinluosl can you update that PR please.

PR #1515, opened 2015-01-27
Add SegmentedBar component

rayshan:
:+1: Ready to be reviewed & merged
===

- [x] Tests - should cover 100% of behavior; all 17 tests pass locally; this PR fixes all the whitespace issues in `run.js`, the only code change in that file is adding the last spec
- [x] Docs
- [x] Demo - serve `index.html` in root and use newly committed `overview` module to review

Questions:

- [x] Need separate controllers or fine to put logic in component object? - no
- [x] Namespace component everywhere, e.g. `ms-segmented-bar`? - yes, `montage-SegmentedBar-whatever`
- [x] Response design - does SegmentedBar need to be vertical at all? Seems questionable. - yes
- [x] Research how select binds user content - done, likely this implementation is more optimal due to conditionally 1-way binding `content` or `contentController`

E.g. for wide desktop:

screen shot 2015-01-26 at 4 48 36 pm

For narrow mobile screen:
screen shot 2015-01-26 at 4 47 58 pm

### DIFF
--- a/test/run.js
+++ b/test/run.js
@@ -95,5 +95,6 @@ require("montage-testing").run(require, [
     "ui/repetition-binding-spec",
     "ui/slot-spec",
     "ui/substitution-spec",
-    "ui/text/text-spec"
+    "ui/text/text-spec",
+    "ui/segmented-bar/segmented-bar-spec"
 ]);
--- a/test/ui/segmented-bar/segmented-bar-spec.js
+++ b/test/ui/segmented-bar/segmented-bar-spec.js
@@ -1,0 +1,103 @@
+// "use strict"; // TODO: causes q to throw, will reinstate when q is replaced by bluebird
+
+var TestPageLoader = require("montage-testing/testpageloader").TestPageLoader;
+
+TestPageLoader.queueTest("segmented-bar-test", function (testPage) {
+    describe("ui/segmented-bar/segmented-bar-spec", function () {
+        var segmentedBarComponent;
+        var segmentElements;
+        beforeEach(function () {
+            segmentedBarComponent = testPage.getElementById("segmentedBar1").component;
+            segmentElements = testPage.querySelectorAll("[data-montage-id=segment1]");
+        });
+
+        afterEach(function () {
+            segmentedBarComponent.exitDocument();
+        });
+
+        describe("SegmentedBar", function () {
+            it("should instantiate with content array", function () {
+                expect(segmentedBarComponent).toBeDefined();
+            });
+
+            it("should instantiate with RangeController", function () {
+                segmentedBarComponent = testPage.getElementById("segmentedBar2").component;
+                expect(segmentedBarComponent).toBeDefined();
+            });
+
+            it("should instantiate with TreeController", function () {
+                segmentedBarComponent = testPage.getElementById("segmentedBar3").component;
+                expect(segmentedBarComponent).toBeDefined();
+            });
+
+            it("should render according to parentElement's dimensions", function () {
+                expect(segmentedBarComponent.element.clientHeight).toEqual(30);
+                expect(segmentedBarComponent.element.clientWidth)
+                    .toEqual(segmentedBarComponent.element.parentElement.clientWidth);
+            });
+
+            it("should repeat the right number of segments", function () {
+                expect(segmentElements.length).toEqual(6);
+            });
+
+            describe("SegmentedBar events", function () {
+                var _handler;
+                beforeEach(function () {
+                    _handler = jasmine.createSpy().andCallFake(function (event) {
+                        expect(event.detail.get("data").label).toEqual("region1");
+                    });
+                    segmentedBarComponent.addEventListener("action", _handler);
+                });
+                afterEach(function () {
+                    segmentedBarComponent.removeEventListener("action", _handler);
+                });
+
+                it("should dispatch an action event with correct .detail data", function () {
+                    testPage.mouseEvent({target: segmentElements[0]}, "mousedown", function () {
+                        testPage.mouseEvent({target: segmentElements[0]}, "mouseup", function () {
+                            expect(_handler).toHaveBeenCalled();
+                        });
+                    });
+                });
+
+                it("should not dispatchActionEvent when enabled === false", function () {
+                    segmentElements = testPage.querySelectorAll("[data-montage-id=segment2]");
+                    testPage.mouseEvent({target: segmentElements[0]}, "mousedown", function () {
+                        testPage.mouseEvent({target: segmentElements[0]}, "mouseup", function () {
+                            expect(_handler).not.toHaveBeenCalled();
+                        });
+                    });
+                });
+            });
+        });
+
+        describe("Segment", function () {
+            it("should instantiate", function () {
+                expect(segmentElements[0].component).toBeDefined();
+            });
+
+            it("should render with correct text label from content array", function () {
+                expect(segmentElements[3].component.templateObjects.text.element.textContent).toEqual("region4");
+            });
+
+            it("should render with correct text label from RangeController", function () {
+                segmentElements = testPage.querySelectorAll("[data-montage-id=segment2]");
+                expect(segmentElements[3].component.templateObjects.text.element.textContent).toEqual("region4");
+            });
+
+            it("should render with correct text label from TreeController", function () {
+                segmentElements = testPage.querySelectorAll("[data-montage-id=segment3]");
+                expect(segmentElements[1].component.templateObjects.text.element.textContent).toEqual("CA");
+            });
+
+            it("should render with correct width in horizontal orientation", function () {
+                expect(segmentElements[0].clientWidth).toEqual(96);
+            });
+
+            it("should render with correct height in vertical orientation", function () {
+                segmentElements = testPage.querySelectorAll("[data-montage-id=segment4]");
+                expect(segmentElements[0].clientHeight).toEqual(96);
+            });
+        });
+    });
+});
--- a/test/ui/segmented-bar/segmented-bar-test.html
+++ b/test/ui/segmented-bar/segmented-bar-test.html
@@ -1,0 +1,146 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>SegmentedBar</title>
+    <script type="text/javascript" src="../../../montage.js" data-package="../../"></script>
+    <script type="text/montage-serialization">
+    {
+        "test": {
+            "prototype": "ui/segmented-bar/segmented-bar-test"
+        },
+        "rangeController": {
+            "prototype": "montage/core/range-controller",
+            "bindings": {
+                "content": {"<-": "@test.data"}
+            }
+        },
+        "treeController": {
+            "prototype": "montage/core/tree-controller",
+            "properties": {
+                "childrenPath": "regions"
+            },
+            "bindings": {
+                "content": {"<-": "@test.tree"}
+            }
+        },
+        "segmentedBar1": {
+            "prototype": "montage/ui/segmented-bar.reel",
+            "properties": {
+                "element": {"#": "segmentedBar1"}
+            },
+            "bindings": {
+                "data": {"<-": "@test.data"}
+            },
+            "listeners": [
+                {
+                    "type": "action",
+                    "listener": {"@": "test"}
+                }
+            ]
+        },
+        "segment1": {
+            "prototype": "montage/ui/segment.reel",
+            "properties": {
+                "element": {"#": "segment1"}
+            },
+            "bindings": {
+                "label": {"<-": "@segmentedBar1:segment.object.region"},
+                "length": {"<-": "@segmentedBar1:segment.object.size"}
+            }
+        },
+        "segmentedBar2": {
+            "prototype": "montage/ui/segmented-bar.reel",
+            "properties": {
+                "element": {"#": "segmentedBar2"},
+                "dataController": {"@": "rangeController"},
+                "enabled": false
+            },
+            "listeners": [
+                {
+                    "type": "action",
+                    "listener": {"@": "test"}
+                }
+            ]
+        },
+        "segment2": {
+            "prototype": "montage/ui/segment.reel",
+            "properties": {
+                "element": {"#": "segment2"}
+            },
+            "bindings": {
+                "label": {"<-": "@segmentedBar2:segment.object.region"},
+                "length": {"<-": "@segmentedBar2:segment.object.size"}
+            }
+        },
+        "segmentedBar3": {
+            "prototype": "montage/ui/segmented-bar.reel",
+            "properties": {
+                "element": {"#": "segmentedBar3"},
+                "dataController": {"@": "treeController"}
+            },
+            "listeners": [
+                {
+                    "type": "action",
+                    "listener": {"@": "test"}
+                }
+            ]
+        },
+        "segment3": {
+            "prototype": "montage/ui/segment.reel",
+            "properties": {
+                "element": {"#": "segment3"}
+            },
+            "bindings": {
+                "label": {"<-": "@segmentedBar3:segment.object.content.region"},
+                "length": {"<-": "@segmentedBar3:segment.object.content.size"}
+            }
+        },
+        "segmentedBar4": {
+            "prototype": "montage/ui/segmented-bar.reel",
+            "properties": {
+                "element": {"#": "segmentedBar4"},
+                "dataController": {"@": "rangeController"},
+                "orientation": "vertical"
+            },
+            "listeners": [
+                {
+                    "type": "action",
+                    "listener": {"@": "test"}
+                }
+            ]
+        },
+        "segment4": {
+            "prototype": "montage/ui/segment.reel",
+            "properties": {
+                "element": {"#": "segment4"}
+            },
+            "bindings": {
+                "label": {"<-": "@segmentedBar4:segment.object.region"},
+                "length": {"<-": "@segmentedBar4:segment.object.size"}
+            }
+        }
+    }
+    </script>
+</head>
+<body>
+    <div style="width: 960px; height: 30px;">
+        <div data-montage-id="segmentedBar1" id="segmentedBar1">
+            <div data-montage-id="segment1"></div>
+        </div>
+    </div>
+
+    <div data-montage-id="segmentedBar2" id="segmentedBar2">
+        <div data-montage-id="segment2"></div>
+    </div>
+
+    <div data-montage-id="segmentedBar3" id="segmentedBar3">
+        <div data-montage-id="segment3"></div>
+    </div>
+
+    <div style="width: 30px; height: 960px;">
+        <div data-montage-id="segmentedBar4" id="segmentedBar4">
+            <div data-montage-id="segment4"></div>
+        </div>
+    </div>
+</body>
+</html>
--- a/test/ui/segmented-bar/segmented-bar-test.js
+++ b/test/ui/segmented-bar/segmented-bar-test.js
@@ -1,0 +1,43 @@
+"use strict";
+
+var TestController = require("montage-testing/test-controller").TestController;
+
+exports.SegmentedBarTest = TestController.specialize({
+    data: {
+        value: [
+            {"region": "region1", "size": 10, "flag": true},
+            {"region": "region2", "size": 10, "flag": true},
+            {"region": "region3", "size": 10, "flag": true},
+            {"region": "region4", "size": 10, "flag": true},
+            {"region": "region5", "size": 10, "flag": true},
+            {"region": "region6", "size": 50, "flag": true}
+        ]
+    },
+
+    tree: {
+        value: {
+            region: "North America",
+            size: 40,
+            regions: [
+                {
+                    region: "Canada",
+                    size: 20
+                },
+                {
+                    region: "United States",
+                    size: 20,
+                    regions: [
+                        {
+                            region: "CA",
+                            size: 40
+                        },
+                        {
+                            region: "MN",
+                            size: 30
+                        }
+                    ]
+                }
+            ]
+        }
+    }
+});
--- a/ui/segment.reel/segment.css
+++ b/ui/segment.reel/segment.css
@@ -1,0 +1,14 @@
+/* autoprefixer > 1%, last 2 versions, Firefox ESR, Opera 12.1 */
+
+.montage-Segment {
+    /* to vertical align text labels */
+    display: -webkit-box;
+    display: -webkit-flex;
+    display: -ms-flexbox;
+    display: flex;
+
+    -webkit-box-align: center;
+    -webkit-align-items: center;
+    -ms-flex-align: center;
+    align-items: center;
+}
--- a/ui/segment.reel/segment.html
+++ b/ui/segment.reel/segment.html
@@ -1,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title></title>
+    <link rel="stylesheet" type="text/css" href="segment.css">
+    <script type="text/montage-serialization">
+    {
+        "owner": {
+            "properties": {
+                "element": {"#": "Segment"}
+            }
+        },
+        "text": {
+            "prototype": "ui/text.reel",
+            "properties": {
+                "element": {"#": "text"}
+            },
+            "bindings": {
+                "value": {"<-": "@owner.label"}
+            }
+        }
+    }
+    </script>
+</head>
+<body>
+    <div data-montage-id="Segment" class="montage-Segment">
+        <span data-montage-id="text"></span>
+    </div>
+</body>
+</html>
--- a/ui/segment.reel/segment.js
+++ b/ui/segment.reel/segment.js
@@ -1,0 +1,74 @@
+"use strict";
+
+/**
+ * @module "ui/segmented-bar.reel/segment.reel"
+ * @requires montage/ui/component
+ */
+var Component = require("ui/component").Component;
+
+/**
+ * @class Segment
+ *
+ * A `Segment` to be used inside a {@link SegmentedBar}.
+ * Each `Segment`'s `length` is a percentage of the entire `SegmentedBar`. Can optional have a `label`.
+ *
+ * @extends Component
+ */
+exports.Segment = Component.specialize( /** @lends Segment.prototype */ {
+    /**
+     * @private
+     */
+    _label: {value: null},
+
+    /**
+     * Text label for each `Segment`. Bound to `Text` component in template.
+     *
+     * @property {String} label - optional
+     */
+    label: {
+        get: function () {return this._label},
+        set: function (label) {
+            if (label !== this._label) {
+                this._label = label;
+                this.needsDraw = true;
+            }
+        }
+    },
+
+    /**
+     * @private
+     */
+    _length: {value: null},
+
+    /**
+     * Used to draw `Segment` length using Flexbox.
+     *
+     * @property {Number} newData.length
+     */
+    length: {
+        get: function () {return this._length},
+        set: function (length) {
+            if (length !== this._length) {
+                this._length = length;
+                this.needsDraw = true;
+            }
+        }
+    },
+
+    /**
+     * Redraw based on updated length.
+     *
+     * @function
+     */
+    draw: {
+        value: function () {
+            if (this.length) {
+                // Autoprefixer > 1%, last 2 versions, Firefox ESR, Opera 12.1
+                this.element.style.webkitBoxFlex = this.length;
+                this.element.style.webkitFlex = this.length;
+                this.element.style.msFlex = this.length;
+                this.element.style.flex = this.length;
+            }
+        }
+    }
+});
--- a/ui/segmented-bar.info/sample/main.reel/main.html
+++ b/ui/segmented-bar.info/sample/main.reel/main.html
@@ -1,0 +1,178 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta http-equiv="content-type" content="text/html; charset=utf-8">
+    <title>Main</title>
+
+    <link rel="stylesheet" type="text/css" href="main.css">
+    <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/bootstrap/latest/css/bootstrap.min.css">
+
+    <script type="text/montage-serialization">
+    {
+        "owner": {
+            "properties": {
+                "element": {"#": "main"}
+            }
+        },
+        "rangeController": {
+            "prototype": "core/range-controller",
+            "bindings": {
+                "content": {"<-": "@owner.data"}
+            }
+        },
+        "treeController": {
+            "prototype": "core/tree-controller",
+            "properties": {
+                "childrenPath": "regions"
+            },
+            "bindings": {
+                "content": {"<-": "@owner.tree"}
+            }
+        },
+        "text": {
+            "prototype": "ui/text.reel",
+            "properties": {
+                "element": {"#": "text"}
+            },
+            "bindings": {
+                "value": {"<-": "@owner.currentSegment"}
+            }
+        },
+        "segmentedBar1": {
+            "prototype": "ui/segmented-bar.reel",
+            "properties": {
+                "element": {"#": "segmentedBar1"}
+            },
+            "bindings": {
+                "data": {"<-": "@owner.data"}
+            },
+            "listeners": [
+                {
+                    "type": "action",
+                    "listener": {"@": "owner"}
+                }
+            ]
+        },
+        "segment1": {
+            "prototype": "ui/segment.reel",
+            "properties": {
+                "element": {"#": "segment1"}
+            },
+            "bindings": {
+                "label": {"<-": "@segmentedBar1:segment.object.region"},
+                "length": {"<-": "@segmentedBar1:segment.object.size"}
+            }
+        },
+        "segmentedBar2": {
+            "prototype": "ui/segmented-bar.reel",
+            "properties": {
+                "element": {"#": "segmentedBar2"},
+                "dataController": {"@": "rangeController"},
+                "enabled": false
+            },
+            "listeners": [
+                {
+                    "type": "action",
+                    "listener": {"@": "owner"}
+                }
+            ]
+        },
+        "segment2": {
+            "prototype": "ui/segment.reel",
+            "properties": {
+                "element": {"#": "segment2"}
+            },
+            "bindings": {
+                "label": {"<-": "@segmentedBar2:segment.object.region"},
+                "length": {"<-": "@segmentedBar2:segment.object.size"}
+            }
+        },
+        "segmentedBar3": {
+            "prototype": "ui/segmented-bar.reel",
+            "properties": {
+                "element": {"#": "segmentedBar3"},
+                "dataController": {"@": "treeController"}
+            },
+            "listeners": [
+                {
+                    "type": "action",
+                    "listener": {"@": "owner"}
+                }
+            ]
+        },
+        "segment3": {
+            "prototype": "ui/segment.reel",
+            "properties": {
+                "element": {"#": "segment3"}
+            },
+            "bindings": {
+                "label": {"<-": "@segmentedBar3:segment.object.content.region"},
+                "length": {"<-": "@segmentedBar3:segment.object.content.size"}
+            }
+        },
+        "segmentedBar4": {
+            "prototype": "ui/segmented-bar.reel",
+            "properties": {
+                "element": {"#": "segmentedBar4"},
+                "dataController": {"@": "rangeController"},
+                "orientation": "vertical"
+            },
+            "listeners": [
+                {
+                    "type": "action",
+                    "listener": {"@": "owner"}
+                }
+            ]
+        },
+        "segment4": {
+            "prototype": "ui/segment.reel",
+            "properties": {
+                "element": {"#": "segment4"}
+            },
+            "bindings": {
+                "label": {"<-": "@segmentedBar4:segment.object.region"},
+                "length": {"<-": "@segmentedBar4:segment.object.size"}
+            }
+        }
+    }
+    </script>
+</head>
+<body>
+    <div data-montage-id="main" class="Main">
+        <h1>Data from event listener</h1>
+        ActionEvent is fired upon press even when data didn't change; please open console to see log of events being fired.
+        <pre data-montage-id="text"></pre>
+
+        <h1>Desktop - data array</h1>
+        <div class="desktop">
+            <div data-montage-id="segmentedBar1">
+                <div data-montage-id="segment1"></div>
+            </div>
+        </div>
+        <br/>
+
+        <h1>Desktop - RangeController + bootstrap, enabled=false</h1>
+        <div class="desktop">
+            <div data-montage-id="segmentedBar2" class="progress">
+                <div data-montage-id="segment2" class="progress-bar progress-bar-striped"></div>
+            </div>
+        </div>
+        <br/>
+
+        <h1>Desktop - TreeController</h1>
+        <div class="desktop">
+            <div data-montage-id="segmentedBar3">
+                <div data-montage-id="segment3"></div>
+            </div>
+        </div>
+        <br/>
+
+        <h1>Phone-portrait, orientation = vertical</h1>
+        <div class="phone-portrait">
+            <div data-montage-id="segmentedBar4">
+                <div data-montage-id="segment4"></div>
+            </div>
+        </div>
+    </div>
+</body>
+</html>
--- a/ui/segmented-bar.info/sample/main.reel/main.js
+++ b/ui/segmented-bar.info/sample/main.reel/main.js
@@ -1,0 +1,56 @@
+"use strict";
+
+var Component = require("ui/component").Component;
+
+exports.Main = Component.specialize({
+    handleAction: {
+        value: function (event) {
+            console.log("heard");
+            if (event.detail) {
+                this.currentSegment = event.detail.get('data').label;
+            }
+        }
+    },
+
+    /**
+     * Data can be any Object[] with any Object structure,
+     * developer will pick right key by binding desired FRB expression.
+     */
+    data: {
+        value: [
+            {"region": "region1", "size": 10, "flag": true},
+            {"region": "region2", "size": 10, "flag": true},
+            {"region": "region3", "size": 10, "flag": true},
+            {"region": "region4", "size": 10, "flag": true},
+            {"region": "region5", "size": 10, "flag": true},
+            {"region": "region6", "size": 50, "flag": true}
+        ]
+    },
+
+    tree: {
+        value: {
+            region: "North America",
+            size: 40,
+            regions: [
+                {
+                    region: "Canada",
+                    size: 20
+                },
+                {
+                    region: "United States",
+                    size: 20,
+                    regions: [
+                        {
+                            region: "CA",
+                            size: 40
+                        },
+                        {
+                            region: "MN",
+                            size: 30
+                        }
+                    ]
+                }
+            ]
+        }
+    }
+});
--- a/ui/segmented-bar.reel/segmented-bar.css
+++ b/ui/segmented-bar.reel/segmented-bar.css
@@ -1,0 +1,20 @@
+/* Autoprefixer > 1%, last 2 versions, Firefox ESR, Opera 12.1 */
+
+/* Horizontal; vertical is set in draw in js */
+.montage-SegmentedBar,
+.montage-SegmentedBar-container {
+    height: 100%;
+}
+
+.montage-SegmentedBar-container {
+    display: -webkit-box;
+    display: -webkit-flex;
+    display: -ms-flexbox;
+    display: flex;
+
+    -webkit-box-direction: normal;
+}
+
+.selected {
+    color: red;
+}
--- a/ui/segmented-bar.reel/segmented-bar.html
+++ b/ui/segmented-bar.reel/segmented-bar.html
@@ -1,0 +1,42 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title></title>
+    <link rel="stylesheet" type="text/css" href="segmented-bar.css">
+    <script type="text/montage-serialization">
+    {
+        "owner": {
+            "properties": {
+                "element": {"#": "SegmentedBar"}
+            }
+        },
+        "repetition": {
+            "prototype": "ui/repetition.reel",
+            "properties": {
+                "element": {"#": "repetition"},
+                "isSelectionEnabled": false
+            },
+            "bindings": {
+                "contentController": {"<-": "@owner._repetitionDataController"}
+            }
+        },
+        ":segment": {
+            "alias": "@repetition:iteration",
+            "comments": [
+                "alias doesn't support @repetition:iteration.data",
+                "see https://github.com/montagejs/montage/issues/1517"
+            ]
+        }
+    }
+    </script>
+</head>
+<body>
+    <div data-montage-id="SegmentedBar" class="montage-SegmentedBar">
+        <!-- Repetition doesn't work on other component's element yet, -->
+        <!-- and we cannot inherit from both Repetition & AbstractButton. -->
+        <div data-montage-id="repetition" class="montage-SegmentedBar-container">
+            <div data-param="*"></div>
+        </div>
+    </div>
+</body>
+</html>
--- a/ui/segmented-bar.reel/segmented-bar.js
+++ b/ui/segmented-bar.reel/segmented-bar.js
@@ -1,0 +1,233 @@
+"use strict";
+
+/**
+ * @module "montage/ui/segmented-bar.reel"
+ * @requires montage/ui/base/abstract-button
+ * @requires montage/core/range-controller
+ * @requires montage/core/tree-controller
+ */
+var AbstractButton = require("ui/base/abstract-button").AbstractButton,
+    RangeController = require("core/range-controller").RangeController,
+    TreeController = require("core/tree-controller").TreeController;
+
+/**
+ * @class SegmentedBar
+ *
+ * A bar consisted of {@link Segment} components. Commonly used to visualize data, such as disk space usage.
+ * Will accept an array of objects, `RangeController` or `TreeController` as data source.
+ * `SegmentedBar` subclasses `AbstractButton` because of `AbstractButton`'s desired `PressComposer` behavior.
+ *
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Flexible_boxes}
+ * for FlexBox usage
+ *
+ * @extends AbstractButton
+ */
+exports.SegmentedBar = AbstractButton.specialize( /** @lends SegmentedBar.prototype */ {
+    /**
+     * Dispatches by default based on the `enabled` property.
+     *
+     * @event SegmentedBar#action
+     * @property {Dict} detail - selected segment's data
+     * @property {String} detail.get("data").label
+     * @property {Number} detail.get("data").length
+     *
+     * @example To access passed data, do `event.detail.get('data')` in event handler
+     */
+
+    /**
+     * @private
+     */
+    _data: {value: null},
+
+    /**
+     * @property {Object[]}
+     * @default null
+     */
+    data: {
+        get: function () {return this._repetitionDataController.content;},
+        set: function (data) {
+            if (data !== this._data) {
+                this._data = data;
+
+                if (data) {
+                    this._dataController = null;
+                    this._repetitionDataController = new RangeController(data);
+                }
+
+                this.needsDraw = true;
+            }
+        }
+    },
+
+    /**
+     * @private
+     */
+    _dataController: {value: null},
+
+    /**
+     * If `TreeController` is used, it will be flattened and only leaves without `children` will be rendered.
+     *
+     * @property {RangeController|TreeController} value
+     * @default null
+     */
+    dataController: {
+        get: function () {return this._dataController;},
+        set: function (dataController) {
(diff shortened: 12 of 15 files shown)


ui/segmented-bar.reel/segment.reel/segment.css
```diff
@@ -0,0 +1,13 @@
...
+    box-sizing: border-box;
+    border: 1px solid black;
+    background: darkgray;
+}
+
+/* vertical, not used, set in JS */
+/*@media (max-width: 1000px) {*/
+/*.montage-Segment {*/
+/*}*/
+/*}*/
```
dharcourt: If this is not used, why is it in the code? Just wondering...

ui/segmented-bar.reel/segment.reel/segment.js
```diff
@@ -0,0 +1,93 @@
...
+ * based on a percentage of the bar's total length. The remaining space
+ * will be its own Segment in the SegmentedBar.
+ */
+exports.Segment = Component.specialize({
+    label: {value: null},
+    _value: {value: null},
+    value: {
+        get: function () {return this._value},
+        set: function (value) {
+            this._value = value;
```
thibaultzanini: You can maybe check if the new value is different from the old one in order to avoid some process.

ui/segmented-bar.reel/segment.reel/segment.js
```diff
@@ -0,0 +1,93 @@
...
+    _calculateLength: {value: function _calculateLength(value) {
+        this._length = value / this.maxValue * 100 + "%";
+    }},
+
+    enterDocument: {value: function () {
+        // set initial segment width
+        this._calculateLength(this.value);
+
+        // TODO: perf review of listening here; doesn't work in prepareForActivationEvents
+        this.addPathChangeListener("orientation", this, "handleOrientationChange");
```
thibaultzanini: You could remove it when the component is not in the dom anymore, exitDocument.

ui/segmented-bar.reel/segment.reel/segment.js
```diff
@@ -0,0 +1,93 @@
...
+            this._calculateLength(value);
+            this.needsDraw = true;
+        }
+    },
+    /**
+     * Length of each segment in %
+     * calculated based on data value & SegmentedBar's maxValue
+     */
+    _length: {value: null},
+    _calculateLength: {value: function _calculateLength(value) {
```
thibaultzanini: Can you indent a little bit more your code, is not really easy to read.

ui/segmented-bar.reel/segment.reel/segment.js
```diff
@@ -0,0 +1,93 @@
...
+     */
+    draw: {value: function () {
+        if (this.orientation === "horizontal") {
+            this.element.style.width = this._length;
+            this.element.style.height = "100%";
+            this.element.style["padding-left"] = "5px";
+        } else {
+            this.element.style.height = this._length;
+            this.element.style.width = "100%";
+            this.element.style["padding-left"] = "0px";
```
thibaultzanini: Can be access directly by the property paddingLeft, it's faster than using the brackets.

ui/segmented-bar.reel/segment.reel/segment.css
rayshan: Sorry should say "TBD, currently not used", b/c we need to decide whether to use media queries in CSS or JS.

ui/segmented-bar.reel/segment.reel/segment.js
```diff
@@ -0,0 +1,93 @@
...
+     * calculated based on data value & SegmentedBar's maxValue
+     */
+    _length: {value: null},
+    _calculateLength: {value: function _calculateLength(value) {
+        this._length = value / this.maxValue * 100 + "%";
+    }},
+
+    enterDocument: {value: function () {
+        // set initial segment width
+        this._calculateLength(this.value);
```
thibaultzanini: Any reason to calculate the length each time the component is added to the Dom, or just the first time should be ok?

ui/segmented-bar.reel/segment.reel/segment.css
dharcourt: If code's not used, I'd rather not see it. Git history's a good place to put stuff you may use later (as long as you don't lose the history), or pastebin, or somewhere else. Think of code as a story or an academic paper: You wouldn't send a story or paper for review, or publish it, with sections commented out and your scratch work included. If you had a question you'd like to raise for the reviewers, you might make a note of it, and the equivalent here would be a single line comment with the question. Commenting out code just makes it invisible to the computer, not to readers, and the secret of beautiful code is to write it to be read and only secondarily be to be executed.

ui/segmented-bar.reel/segment.reel/segment.js
```diff
@@ -0,0 +1,93 @@
...
+        // bubble a CustomEvent & stop propagating heard event
+        event.stopPropagation();
+
+        // TODO: need more data?
+        var _data = {label: this.label, value: this.value};
+        var segmentClickEvent = new CustomEvent(
+            "SegmentClick",
+            {detail: _data, bubbles: true, cancelable: true}
+        );
+        this.element.dispatchEvent(segmentClickEvent);
```
thibaultzanini: You could use our own API for that (dispatchEventNamed) https://github.com/montagejs/montage/blob/master/core/target.js#L122

ui/segmented-bar.reel/segmented-bar.js
```diff
@@ -0,0 +1,132 @@
...
+     */
+    orientation: {value: null},
+    /**
+     * Allows user to override orientation.
+     * TODO: needed?
+     */
+    orientationOverride: {value: null},
+    /**
+     * TODO: needed?
+     */
```
dharcourt: It would be simpler if the API to set the orientation would be the "orientation" property and if there were no "orientationOverride" property.

ui/segmented-bar.reel/segmented-bar.js
```diff
@@ -0,0 +1,132 @@
...
+                console.warn(
+                    "SegmentedBar content values total " + sum +
+                    ", maxValue overridden from " + this.maxValue +
+                    " to match.");
+                this.maxValue = sum;
+            } else {
+                this.remainingValue = this.maxValue - sum;
+
+                // construct remaining segment
+                var remainingSegment = {};
```
marchant: I thought we had agreed that all values are expected to be in the array of data passed to the segmentbar so that we wouldn't have to deal with that, make the contract simpler and consitent, no?

ui/segmented-bar.reel/segment.reel/segment.css
rayshan: Agreed, will fix.

ui/segmented-bar.reel/segmented-bar.js
rayshan: I have in my notes current implementation is what we discussed, not sure if notes came from you or @dharcourt though. Which of these would we prefer?

"As a user, I'll pass the disk capacity and current disk usage, for a segmented bar to be constructed."

"As a user, I'll pass the current disk usage, since I don't know the remaining disk space, I'll subtract current disk usage out of disk capacity, then pass the result, for a segmented bar to be constructed."

I would think the first scenario puts less work on the user. If user doesn't know disk capacity, she doesn't need to pass it, since `maxValue` is optional.

@dharcourt?

dharcourt: The user story text above is confusing... Is this from the point of view of a developer using the SegmentedBar, as suggested by the context, or from the point of someone using an application that has a SegmentedBar, as suggested by the "As a user" text? Also, this looks a bit like the [format of user stories](http://en.wikipedia.org/wiki/User_story#Format), but it's actually almost the opposite (stating what is required of the interested party, not what is desired by them).

Anyway, I'm going to guess that what we're discussing here is whether the SegmentedBar should have a "leftover value" segment specified through a "totalValue" or "total" property. The term "maxValue" would be wrong for this, by the way, because it is actually not the maximum of any value provided by the API caller. It could be considered the maximum of the total of values provided by API callers, but even that's not clear because that depends on what happens when the total of all value exceeds this "total" value... Oh the confusion!

I don't know yet if we have a need for a leftover segment in Contour. In the absence of a known need, and in the interest of simplicity, I would strongly suggest we not support a leftover segment for now in SegmentedBar. It's not clear whether we should ever support it, because: 1) A leftover can alway be represented as an extra element in the data passed to a leftover-less SegmentedBar; and 2) It makes everything so confusing, changing "a bar split into segments" into "a bar that represents an amount, which sub-amounts represented as segments on that bar and leftover value represented as a different type of segment, and somewhat undefined behavior if the total of the sub-amounts exceeds the bar's amount". The later may be a use case of SegmentedBar, but it would be easier for everyone including ourselves to keep the definition of the SegmentedBar as the former :-).

By the way, here's how easy it would be to represent a leftover in data passed in to the simpler version of SegmentedBar... if that's ever needed:
```
segmentedBarData: {
    get: function() {
        return {children: [this.valueTree, {value: this.capacity - this.valueTree.valueTotal}]}
    }
}
```

ui/segmented-bar.reel/segment.reel/segment.js
rayshan: Thanks will implement.

ui/segmented-bar.reel/segment.reel/segment.js
rayshan: Thanks will put value: on new line which will help w/ indent

ui/segmented-bar.reel/segment.reel/segment.js
rayshan: Thanks will implement.

ui/segmented-bar.reel/segment.reel/segment.js
rayshan: Thanks will implement

ui/segmented-bar.reel/segment.reel/segment.js
rayshan: Good point, since we're switching to flexbox, this is no longer needed.

ui/segmented-bar.reel/segment.reel/segment.js
rayshan: Learned something new! Thanks!

ui/segmented-bar.reel/segmented-bar.js
rayshan: Thanks will refactor.

ui/segmented-bar.reel/segmented-bar.js
rayshan: Thanks for the notes. Sorry you're correct "user story" is more how a developer would use the component. I'll try to get better at this. For now we'll remove support for remaining segments.

rayshan:
Ping, :+1: ready to be reviewed & merged

ui/segment.reel/segment.css
```diff
@@ -0,0 +1,12 @@
...
+    display: -webkit-box;
+    display: -webkit-flex;
+    display: -ms-flexbox;
+    display: flex;
+
+    -webkit-box-align: center;
+    -webkit-align-items: center;
+    -ms-flex-align: center;
+    align-items: center;
+}
```
dharcourt: I think -webkit-box is needed for Android browser 4.3... are we supporting Android browsers before 4.4?

rayshan: Prefixes are chosen by using https://github.com/postcss/autoprefixer, with default browser choise of `> 1%, last 2 versions, Firefox ESR, Opera 12.1`, more details here: https://github.com/ai/browserslist

ui/segment.reel/segment.html
```diff
@@ -0,0 +1,30 @@
...
+        }
+    }
+    </script>
+</head>
+<body>
+<div data-montage-id="Segment" class="montage-Segment">
+    <span data-montage-id="text"></span>
+</div>
+</body>
+</html>
```
dharcourt: nit: why is the <head> contents indented but not the <body> ones? It'd be both more consistent and more readable to have the <body> contents indented too.

ui/segment.reel/segment.js
```diff
@@ -0,0 +1,66 @@
...
+     */
+    draw: {value: function () {
+        if (this.content) {
+            this.element.style.webkitBoxFlex = this.content.length;
+            this.element.style.webkitFlex = this.content.length;
+            this.element.style.msFlex = this.content.length;
+            this.element.style.flex = this.content.length;
+        }
+    }}
+});
```
dharcourt: I thought we'd agreed to switch from "content" to "data" for APIs going forward... no?

ui/segment.reel/segment.html
rayshan: It's a webstorm thing, this way things inside body isn't indented too deeply, will fix

ui/segment.reel/segment.js
rayshan: This is what we agreed on:

> content to data; contentController to dataController; iteration.object to iteration.data

Should we start changing now or wait for Repetition to be changed first? If we start now we still have to use iteration.object. I'm ok either way. Your call.

ui/segment.reel/segment.css
dharcourt: Fair enough... kind of. I'm not thrilled about a tool deciding which browsers our code supports, especially when someone reading the code isn't going to know what tool generated it, what our settings were for this tool, and generally why it generated the values it generated. Can we configure autoprefixer to just generate code for the browsers we support, like a human would have done? What happens if we're committed to supporting browsers that autoprefixer decides to drop because of dwindling market share?

In any case, it may be worth putting a comment in this .css files, even if it's just one at the top explaining that autoprefixer was used to generate the browser variant prefixes. Regardless of what tool is used, we humans are responsible for making sure the code we commit isn't generating too many WTF's (like my WTF is -webkit-box doing here).

ui/segment.reel/segment.js
dharcourt: `data` and `dataController` as the main properties for SegmentedBar, `content` and `contentController` as covers (with setter/getters that get the corresponding values and log a gentle "deprecated" message to the console). We'll live with `iteration.object` for now and until Repetition is updated similarly, but we might as well do the update work for SegmentedBar now since we're committing that code.

ui/segmented-bar.info/main.reel/main.html
```diff
@@ -0,0 +1,180 @@
...
+    <h1>Phone-portrait, orientation = vertical</h1>
+    <div class="phone-portrait">
+        <div data-montage-id="segmentedBar4">
+            <div data-montage-id="segment4"></div>
+        </div>
+    </div>
+</div>
+
+</body>
+</html>
```
dharcourt: nit: MontageStudio indents the serialization (lines 11-137) at the same level as the enclosing script. Manually edited serialization should have the same indentation (or MontageStudio should be changed if there's a good reason to change its indentation).

dharcourt: nit: <body> contents should be indented.

ui/segmented-bar.info/sample/main.reel/main.html
```diff
@@ -0,0 +1,180 @@
...
+    <h1>Phone-portrait, orientation = vertical</h1>
+    <div class="phone-portrait">
+        <div data-montage-id="segmentedBar4">
+            <div data-montage-id="segment4"></div>
+        </div>
+    </div>
+</div>
+
+</body>
+</html>
```
dharcourt: nit: <body> contents should be indented.

ui/segmented-bar.reel/segmented-bar.html
```diff
@@ -0,0 +1,40 @@
...
+
+<div data-montage-id="SegmentedBar" class="montage-SegmentedBar">
+    <!--repetition doesn't work on other component's element yet-->
+    <div data-montage-id="repetition" class="montage-SegmentedBar-container">
+        <div data-param="*"></div>
+    </div>
+</div>
+
+</body>
+</html>
```
dharcourt: nit: `<body>` contents should be indented.

test/run.js
```diff
@@ -1,104 +1,100 @@
...
+    "ui/modal-overlay-spec",
+    "ui/overlay-spec",
+    "ui/repetition-spec",
+    "ui/repetition-selection-spec",
+    "ui/repetition-binding-spec",
+    "ui/slot-spec",
+    "ui/substitution-spec",
+    "ui/text/text-spec",
+    "ui/segmented-bar/segmented-bar-spec"
 ]);
```
dharcourt: Thank you thank you for the cleanup!

test/ui/segmented-bar/segmented-bar-test.html
```diff
@@ -0,0 +1,144 @@
...
+</div>
+
+<div style="width: 30px; height: 960px;">
+    <div data-montage-id="segmentedBar4" id="segmentedBar4">
+        <div data-montage-id="segment4"></div>
+    </div>
+</div>
+
+</body>
+</html>
```
dharcourt: nit: `<body>` contents should be indented, allowing the blank lines around that `<body>` contents to be removed.

dharcourt: "rangeController" is present twice in the serialization.

dharcourt: nit: I'm not sure manually formatting the serialization is useful. MontageStudio formats the serialization very predictably with the @owner first (@test would be the equivalent in this case) and all the other objects serialized by key alphabetical order after that. This makes it easy to find objects, and I'd rather not have to wonder whether Studio or a human formatted a particular source's serializations before trying to find an object, especially if in the later case I have to reverse engineer that human's formatting rationale and wonder for example if the double blank lines mean something special or are just that human's idea of what beautiful code is. Also, having a serialization style different from Studio's may cause unnecessary diffs when code is passed between a Studio user and a manual editor.

ui/segmented-bar.info/main.reel/main.html
dharcourt: nit: MontageStudio-style organization of the serialization (@owner, then alphabetical) would make it easier to read (because more predictable and less personal) and would avoid the double blank lines (lines 17-18, 57-58, 83-84, 110-111) which introduce too much vertical space.

ui/segment.reel/segment.html
dharcourt: I like how the serialization is formatted exactly as MontageStudio would have formatted it :-). Sometimes the little things in life...

test/ui/segmented-bar/segmented-bar-spec.js
```diff
@@ -0,0 +1,110 @@
...
+                expect(segmentElements[0].clientWidth).toEqual(96);
+            });
+
+            it("should render with corrent height in vertical orientation", function () {
+                segmentElements = testPage.querySelectorAll("[data-montage-id=segment4]");
+                expect(segmentElements[0].clientHeight).toEqual(96);
+            });
+        });
+    });
+});
```
dharcourt: nit: The `// teardown to not affect future tests` comments look funky. They're a little long and a little too much on their own line to be sentence fragment comments without capitalization and periods. Perhaps `// Tear down to not affect future tests.` (full sentence on a line by itself, with correct punctuation) or `segmentedBarComponent.exitDocument(); // teardown` (dependent clause describing the line it's on).

test/ui/segmented-bar/segmented-bar-test.js
```diff
@@ -0,0 +1,32 @@
...
+                regions: [
+                    {
+                        region: "China",
+                        size: 40
+                    }
+                ]
+            }
+        ]
+    }}
+});
```
dharcourt: nit: That JSON formatting looks strange: Why is value on the same line as its parent on both places it appears? For that matter, why is there no blank line between the property descriptors, unlike every other `specialize()` call in this pull request?

ui/segment.reel/segment.js
dharcourt: nit: This would be more readable with both JSON "value" occurrences on lines of their own. I may be missing something here but I don't see what JSON formatting style this is.

dharcourt: nit: Why @prop instead of @property? It's an abbreviation, and @property is the primary keyword for this anyway (http://usejsdoc.org/tags-property.html).

dharcourt: It took me about 10 minutes to figure out that "content" was a cover for a length and a label property. Why are these not separate properties? It would require two bindings instead of one, but isn't that the way object properties are usually handled? The ``content`` binding in the JSDoc is not for the faint of heart...

dharcourt: nit: Why `newContent` as the setter argument name? Is using `new` + property name a general style we should switch to for setter arguments? Most of the existing code uses `value

dharcourt: Why is the `content` property marked as `@returns {Object[]}` in the JSDoc comment? Typo perhaps?

dharcourt: nit: E.g. can usually be replaced by a more easily understandable non-abbreviated non-latin equivalent, like "for example" (though I for one have a hard time keeping e.g.'s and i.e.'s out of my email messages :-). This sentence would read better if it started simply with "Example", and is in fact grammatically incorrect as it is now with "e.g.".

ui/segmented-bar.info/main.reel/main.js
```diff
@@ -0,0 +1,47 @@
...
+                                size: 40
+                            }
+                        ]
+                    }
+                ]
+            };
+
+        }
+    }
+});
```
dharcourt: nit: Why is currentSegment all on one line, with no line separating it from the following function? It may save space but the inconsistency slows down reading.

dharcourt: nit: The sample tree data should be made as inoffensive as possible: Canadian developers might not be happy to learn that they're a region of the U.S., or Chinese developers to learn that China is just 40% of the size of the United States. Oh, and it's U.S. or US, not USA, according to http://english.stackexchange.com/questions/159018/the-usa-vs-the-us.

ui/segmented-bar.info/sample/main.reel/main.html
dharcourt: nit: By now, you must know what I think of carefully crafted developer-specific serialization formatting :-).

ui/segmented-bar.info/main.reel/main.html
dharcourt: Why is `Phone-portrait` "phone"-y, and the other examples "desktop"-y? Is there something that would prevent desktop segmented bars from being vertical? Is there something special in this example that relates to touch? If neither of those are true this would be more understandable as simply "vertical" and "horizontal".

dharcourt: Why does the vertical segmented bar have labels? This does not look good, and we shouldn't be giving examples of what we don't want developers to do :-). On the contrary, having an example of a vertical segmented bar without labels would not only correspond to something developers might want to do but also provide an example of how to do it (it may not be obvious to developers how to not have labels).

dharcourt: In general, the examples could be made more pretty, with padding, colors, and the like. There are good reasons why the Montage component doesn't have styling but no good reasons why the example doesn't, particularly since if we provide some nice styling developers might actually be able to copy  and adapt it.

dharcourt: Can we have an example (or all examples) with differently colored segments? This will probably be the usual use case for segmented bar, so we should show how it's done. This will also provide us with verification that it's as easy to do as it should be.

ui/segmented-bar.info/main.reel/main.js
dharcourt: Oops, why is there a segmented-bar.info/main.reel/main.js? Shouldn't it be segmented-bar.info/sample/main.reel/main.js? No wonder I was getting confused about which bit of code I'd commented on already... there are duplicates!

ui/segmented-bar.info/sample/main.reel/main.html
dharcourt: See the comments for ui/segmented-bar.info/main.reel/main.html, they all apply here.

ui/segmented-bar.info/sample/main.reel/main.js
```diff
@@ -0,0 +1,49 @@
...
+                regions: [
+                    {
+                        region: "China",
+                        size: 40
+                    }
+                ]
+            }
+        ]
+    }}
+});
```
dharcourt: See the comments for ui/segmented-bar.info/main.reel/main.js, they all apply here.

ui/segmented-bar.reel/segmented-bar.css
```diff
@@ -0,0 +1,17 @@
...
+    display: -webkit-box;
+    display: -webkit-flex;
+    display: -ms-flexbox;
+    display: flex
+}
+
+.selected {
+    color: red;
+}
+
```
dharcourt: nit: Blank line at the end of the file?

ui/segmented-bar.reel/segmented-bar.js
```diff
@@ -0,0 +1,259 @@
...
+            if (this.orientation === "vertical") {
+                this.templateObjects.repetition.element.style.webkitBoxOrient = "vertical";
+                this.templateObjects.repetition.element.style.webkitBoxDirection = "normal";
+                this.templateObjects.repetition.element.style.webkitFlexDirection = "column";
+                this.templateObjects.repetition.element.style.msFlexDirection = "column";
+                this.templateObjects.repetition.element.style.flexDirection = "column";
+            }
+        }
+    }
+});
```
dharcourt: "OS X storage space visualization found in iTunes or About This Mac" was **not** our "reference design". Otherwise that would be copyright infringement. Don't mess with that stuff. I am assuming you did not slavishly copy the iTunes or Mac OS storage representation.

In any case, there's no reason to mention iTunes or About this Mac as there are plenty of prior examples of segmented bars and we shouldn't be singling out this particular example.

ui/segment.reel/segment.js
dharcourt: nit: Periods missing from the sentences on lines 26 and 27. Does this even look reasonable in JSDoc or does JSDoc run these sentences together?

ui/segmented-bar.reel/segmented-bar.js
dharcourt: nit @ line 11: "A segmented bar" is not a very useful description of SegmentedBar. See "Add description beyond the API name" of http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#styleguide for some ideas.

dharcourt: nit: Missing periods at the ends of the sentences of lines 12, 13, 71, 78, 86, ... and basically everywhere else in this file.

dharcourt: If we can't avoid `childrenPath`, then it should be named `childrenExpression`.

dharcourt: nit @ line 94: "perf"? Is that supposed to refer to something specific (a perf program or something) or is it an abbreviation? If the former, which is that not explained, if the later, why do we have an unexplained abbreviation in documentation?

dharcourt: nit @ line 100: Why is this function named?

dharcourt: nit: e.g. again.

dharcourt: nit @ line 119: Why have single line ifs? They save only a couple of lines at the cost the code harder to read and making it more complicated to add statements in the body of the if. They also promote bad code since it makes it easier to have many different ifs in a method, which is usually a result of bad method factoring or overly complex logic (but not in this case :-).

dharcourt: nit: Many implementation comments like `// prepareForActivationEvents isn't soon enough for this to take effect` lack capitalization and periods. Even though only developers will read those, developers are human beings too and deserve proper spelling and grammar when someone is trying to communicate with them through comments.

dharcourt: nit: Speaking of humans that deserve proper spelling and grammar, ``performant`` isn't a word (yet), and ``Performs better than this.super()`` would be so much nicer to read than ``more performant than this.super()`` on line 172.

dharcourt: line 119: Are we using a private variable of Component? If so, why?

dharcourt: nit @ line 135: Why not a proper english sentence here?

dharcourt: line 138: Why isn't this done through a binding in the serialization?

dharcourt: Why is `_treeControllerClass` set lazily? Is this for performance? In any case a comment about this where `_treeControllerClass` is defined would help.

dharcourt: line 148: What if `childrenPath` (or more properly `childrenExpression`) is set in the TreeController but not in the SegmentedBar? Won't this clobber that set value?

dharcourt: line 165: This comment is not clear at all: `Need to re-enable super's event dispatching b/c overriding`. In general the comments in this file are sloppy, unclear, and disagreeable to read. I'm sure that's not the intention but one gets the impression the author couldn't be bothered to write proper sentences and doesn't really care about the comfort of anyone reading this code.

dharcourt: What if the value of isSelectionEnabled is changed while the SegmentedBar is in the DOM? Will the selection bindings be set or cleared accordingly? It doesn't seem like it.

dharcourt: line 184: `Ensure not to cancel bindings SegmentedBar didn't define`... how is the code doing that? If it's not doing anything special, why is this comment there? This is confusing.

dharcourt: BUG: Line 211 will not dispatch a second action event when the user clicks on a segment, then clicks on that same segment again at a later time but before clicking on another segment. In general it is not possible to use a repetition selection binding to trigger a `dispatchActionEvent()`.

dharcourt: BUG? Will the logic at line 250 work if the orientation is changed from horizontal to vertical, then back to horizontal? I haven't tested, but looking at this code it seems it won't work as expected...

dharcourt: Declaring the `_content` and `_contentController` properties is unnecessarily verbose. Those private properties are only used in the `content` and `contentController` setters and getters, and they don't need to be declared outside of those. As an added bonus, keeping `_content` and `_contentController` to within the `content` and `contentController` accessors will ensure those properties are only created if they are needed.

dharcourt: line 23: What? Why would an action event be tied to a selection flag? I might want to handle segmented bar clicks without being interested in the selection. Also, why is there an isSelectionEnabled flag in the first place? Are we trying to improve performance in those cases where the develop does not want to handle clicks and/or selection? How much time does this save? Is it worth adding an extra flag and concept that the developer wouldn't otherwise have to deal with?

dharcourt: line 110: I think this is wrong... it seems that this will put in the event the "content" object of the selected segment, which is simply a cover for the segment's length and label properties when the segment is a Segment, and may be nothing or something entirely different when the segment is a custom class. What should be in the event is the Repetition:iterator.object corresponding to the segment clicked on.

dharcourt: Note that the whole selection binding set up in `_defineSelectionBinding()` is not necessary: The fact that a user has clicked on the segmented bar needs to be detected (which can't be done through a selection binding), and at that point the SegmentedBar can easily ask its repetition for selection[0] (or more exactly the data corresponding to that), so there's no need for a binding to get the appropriate data into the action event.

dharcourt: BUG (I think): If you set a `dataController` value (legacy name: `contentController` :-), then let the segmented bar draw itself, then set a `data` value (legacy name: `content`), I don't think the segmented bar will correctly update the bindings to the repetition so the new data is displayed (because the bindings are set in templateDidLoad, which is called once and then never again). The most robust way to fix this is to ensure the segmented bar always has an RangeController, either taken or derived from the `dataController` or derived from the `data`. The repetition can then have a binding to `this arrayController` defined in the serialization and the binding doesn't need to change. Setters for `data` and `dataController` will ensure any changes get reflected in a correct `_rangeController`, which will update the repetition. Because you'll have better separation of concerns, the code for this might actually be smaller than the current version. No _data and _dataController properties will be necessary, and It's likely no `needsDraw = 1` will be needed in the `data` and `dataController` setters because the binding will cause changes to `_rangeController` to propagate correctly to the repetition and force a redraw there.

dharcourt: It doesn't seem right that SegmentedBar should have a `childrenPath`/`childrenExpression` that's redundant with TreeController's: If developers want to create a view on a tree that doesn't have the default structure, they can and should create a TreeController and set its childrenExpression.

ui/segment.reel/segment.js
rayshan: > data and dataController as the main properties for SegmentedBar, content and contentController as covers (with setter/getters that get the corresponding values and log a gentle "deprecated" message to the console). We'll live with iteration.object for now and until Repetition is updated similarly, but we might as well do the update work for SegmentedBar now since we're committing that code.

will fix

ui/segment.reel/segment.css
dharcourt: Sounds good. Note that this is not about what our developers will do/use (Studio or mop could indeed handle that), but about the fact that we're responsible for the code we commit, regardless of how that code was written or generated. With a comment, this code will be self-explanatory and the -webkit-box will be fine.

ui/segment.reel/segment.js
rayshan: > nit: This would be more readable with both JSON "value" occurrences on lines of their own. I may be missing something here but I don't see what JSON formatting style this is.

will fix

rayshan: > nit: Why @prop instead of @property? It's an abbreviation, and @property is the primary keyword for this anyway (http://usejsdoc.org/tags-property.html).

Laziness. We have significant inconsistency in JSDoc tag names, see #1540 where I fixed a ton of them. I'll eventually use ESLint to catch JSDoc issues. For now won't fix.

rayshan: > It took me about 10 minutes to figure out that "content" was a cover for a length and a label property. Why are these not separate properties? It would require two bindings instead of one, but isn't that the way object properties are usually handled? The content binding in the JSDoc is not for the faint of heart...

I thought we discussed this and Benoit wanted the user to use FRB expressions. Plus this saves 1 binding per segment * n segments, could be significant. Won't fix.

rayshan: > nit: Why newContent as the setter argument name? Is using new + property name a general style we should switch to for setter arguments? Most of the existing Montage code uses value for that. That's not my favorite, and I prefer code that provides some typing information in the setter argument name (set: function (flag) {... or set: function(string) {... for example), but at least it's consistent and makes it easy to see in the setter where the set value is. new + property name also provides the false impression that the passed in value will be the result of a new X() expression.

No it's not common in this code base. Good point. Will change arg to `data`.

rayshan: > Why is the content property marked as @returns {Object[]} in the JSDoc comment? Typo perhaps?

Not clear on how to document getters & setters, let's wait on https://github.com/jsdoc3/jsdoc/issues/925

rayshan: > nit: E.g. can usually be replaced by a more easily understandable non-abbreviated non-latin equivalent, like "for example" (though I for one have a hard time keeping e.g.'s and i.e.'s out of my email messages :-). This sentence would read better if it started simply with "Example", and is in fact grammatically incorrect as it is now with "e.g.".

will fix

rayshan: > nit: Periods missing from the sentences on lines 26 and 27. Does this even look reasonable in JSDoc or does JSDoc run these sentences together?

JSDoc does not insert new line characters in this case. Will fix.

ui/segmented-bar.info/main.reel/main.html
rayshan: > nit: <body> contents should be indented and the blank lines around it removed.

will fix

rayshan: > nit: MontageStudio-style organization of the serialization (@owner, then alphabetical) would make it easier to read (because more predictable and less personal) and would avoid the double blank lines (lines 17-18, 57-58, 83-84, 110-111) which introduce too much vertical space.

In this codebase serialization objects aren't alphabetical, sometimes even `@owner` isn't at the top. I'll rearrange to owner > range > tree > text > segments, b/c if alphabetical, tree & text will get buried under all the segments.

Will remove double blank lines.

> Also, MontageStudio indents the serialization (lines 11-137) at the same level as the enclosing script... any reason to not do that here?

Indents in html in this codebase are all over the place, but sure, will fix

ui/segmented-bar.reel/segmented-bar.html
rayshan: Will fix all

test/run.js
rayshan: Np!

test/ui/segmented-bar/segmented-bar-test.html
rayshan: > nit: <body> contents should be indented, allowing the blank lines around that <body> contents to be removed.

will fix

> "rangeController" is present twice in the serialization.

oops will fix

> nit: I'm not sure manually formatting the serialization is useful. MontageStudio formats the serialization very predictably with the @owner first (@test would be the equivalent in this case) and all the other objects serialized by key alphabetical order after that. This makes it easy to find objects, and I'd rather not have to wonder whether Studio or a human formatted a particular source's serializations before trying to find an object, especially if in the later case I have to reverse engineer that human's formatting rationale and wonder for example if the double blank lines mean something special or are just that human's idea of what beautiful code is. Also, having a serialization style different from Studio's may cause unnecessary diffs when code is passed between a Studio user and a manual editor.

will fix

test/ui/segmented-bar/segmented-bar-spec.js
rayshan: Moved duplicated code to `afterEach` block

test/ui/segmented-bar/segmented-bar-test.js
rayshan: to save space & indent, will fix

ui/segmented-bar.info/main.reel/main.js
rayshan: > nit: Why is currentSegment all on one line, with no line separating it from the following function? It may save space but the inconsistency slows down reading.

will fix

> nit: The sample tree data should be made as inoffensive as possible: Canadian developers might not be happy to learn that they're a region of the U.S., or Chinese developers to learn that China is just 40% of the size of the United States. Oh, and it's U.S. or US, not USA, according to http://english.stackexchange.com/questions/159018/the-usa-vs-the-us.

changed to ABCD

> Oops, why is there a segmented-bar.info/main.reel/main.js? Shouldn't it be segmented-bar.info/sample/main.reel/main.js? No wonder I was getting confused about which bit of code I'd commented on already... there are duplicates!

That was dumb, sorry! will fix

ui/segmented-bar.info/sample/main.reel/main.html
rayshan: will fix

ui/segmented-bar.info/sample/main.reel/main.js
rayshan: will fix

ui/segmented-bar.reel/segmented-bar.css
rayshan: weird, will fix

ui/segmented-bar.reel/segmented-bar.js
rayshan: > "OS X storage space visualization found in iTunes or About This Mac" was not our "reference design". Otherwise that would be copyright infringement. Don't mess with that stuff. I am assuming you did not slavishly copy the iTunes or Mac OS storage representation. In any case, there's no reason to mention iTunes or About this Mac as there are plenty of prior examples of segmented bars and we shouldn't be singling out this particular example.

good call, will fix

rayshan: > nit @ line 11: "A segmented bar" is not a very useful description of SegmentedBar. See "Add description beyond the API name" of http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#styleguide for some ideas.

Good link. I'll try.

rayshan: > nit: Missing periods at the ends of the sentences of lines 12, 13, 71, 78, 86, ... and basically everywhere else in this file.

As you can see I like "sentence fragments" b/c I write a ton of bulleted lists. I'll see what I can do.

rayshan: > If we can't avoid childrenPath, then it should be named childrenExpression.

> line 148: What if childrenPath (or more properly childrenExpression) is set in the TreeController but not in the SegmentedBar? Won't this clobber that set value?

> It doesn't seem right that SegmentedBar should have a childrenPath/childrenExpression that's redundant with TreeController's: If developers want to create a view on a tree that doesn't have the default structure, they can and should create a TreeController and set its childrenExpression.

That's a good argument. Will remove `childrenPath` from API.

rayshan: > nit @ line 94: "perf"? Is that supposed to refer to something specific (a perf program or something) or is it an abbreviation? If the former, which is that not explained, if the later, why do we have an unexplained abbreviation in documentation?

will fix

rayshan: > nit @ line 100: Why is this function named?

Because it's the only function I created that doesn't overriding parent's. I can't help that parents' functions aren't named, but I can ensure mine are.

rayshan: > nit @ line 113: "e.g. via use of substitution." => "for example when it is used in a Substitution." The "e.g. via" form can actually mean something quite different than what is intended.

will fix

rayshan: > nit @ line 119: Why have single line ifs? They save only a couple of lines at the cost the code harder to read and making it more complicated to add statements in the body of the if. They also promote bad code since it makes it easier to have many different ifs in a method, which is usually a result of bad method factoring or overly complex logic (but not in this case :-).

Good point, will fix, but they're elsewhere in the codebase :(

rayshan: > nit: Many implementation comments like // prepareForActivationEvents isn't soon enough for this to take effect lack capitalization and periods. Even though only developers will read those, developers are human beings too and deserve proper spelling and grammar when someone is trying to communicate with them through comments.

Ok ok I'll try, but can't capitalize `prepareForActivationEvents`.

rayshan: > nit: Speaking of humans that deserve proper spelling and grammar, performant isn't a word (yet), and Performs better than this.super() would be so much nicer to read than more performant than this.super() on line 172.

will fix

rayshan: > line 119: Are we using a private variable of Component? If so, why?

Patch is coming to make `_preparedForActivationEvents` public. We have to check it because `prepareForActivationEvents` is only ever called once, even after `exitDocument`, as discussed 2 weeks ago w/ the team. However upon first `enterDocument` we don't want to define bindings for performance reasons. See https://github.com/montagejs/montage/issues/1521.

dharcourt: You can work around #1521 without introducing a dependency on a private variable by creating your own instance variable in SegmentedBar that tracks whether `prepareForActivationEvents()` was called. You can even take this opportunity to name the variable something that will be more understandable in this context, like `wasSelectionBound` or something. Having a comment referring to #1521 is a very good idea so the temporary variable can be removed in time.

Using `_preparedForActivationEvents` will necessarily introduce a bug when that private variable is later renamed, so it's better to have #1521 cost the space for an extra boolean in every SelectedBar.

Except in very extreme circumstances (which this is not) using private instance variables and private APIs is just plain laziness, often done by people who couldn't think of a better way to solve the problem they faced. You're better than that :-).

ui/segmented-bar.reel/segmented-bar.js
```diff
@@ -0,0 +1,259 @@
...
+ * A segmented bar.
+ * Reference design: OS X storage space visualization found in iTunes or About This Mac
+ * Inheriting from AbstractButton because SegmentedBar needs its PressComposer behavior
+ *
+ * @extends AbstractButton
+ *
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Flexible_boxes}
+ * for FlexBox usage
+ */
+exports.SegmentedBar = AbstractButton.specialize( /** @lends SegmentedBar.prototype # */ {
```
dharcourt: Question: Does it make a difference if there's a space before the @lends pound (`/** @lends SegmentedBar.prototype # */` instead of `/** @lends SegmentedBar.prototype# */`)? I was figuring out how to style my JSDoc comments and came across this variation.

rayshan: Actually I don't know why there's a `#` at all, they're all over the place in the codebase, we need to fix it. 

Though it's crucial to have `.prototype` because they're instance methods/properties, not static.

rayshan: > nit @ line 135: Why not a proper english sentence here?

will fix

rayshan: > line 138: Why isn't this done through a binding in the serialization?

b/c why spend resources on a binding data if user is just passing a dataController

rayshan: > Why is _treeControllerClass set lazily? Is this for performance? In any case a comment about this where _treeControllerClass is defined would help.

`require()` is done lazily for performance; I was doing 2 instance checks in 2 different places, hence the private property, but not anymore, will remove

dharcourt: Duh. Sorry about that.

   - C=

rayshan: > line 165: This comment is not clear at all: Need to re-enable super's event dispatching b/c overriding. In general the comments in this file are sloppy, unclear, and disagreeable to read. I'm sure that's not the intention but one gets the impression the author couldn't be bothered to write proper sentences and doesn't really care about the comfort of anyone reading this code.

Sorry, will fix

rayshan: > What if the value of isSelectionEnabled is changed while the SegmentedBar is in the DOM? Will the selection bindings be set or cleared accordingly? It doesn't seem like it.

There's no logic to account for changing `isSelectionEnabled` after instantiation. I was thinking about the possible use cases for SegmentedBar. It's highly unlikely someone would do that IMO.

I also didn't account for someone setting `data`, then setting `dataController` with different dataset, then go back and force on `data` and `dataController` with different datasets. Also too edge of a case IMO.

How edge must a case be before we decide to not account for it?

rayshan: > line 184: Ensure not to cancel bindings SegmentedBar didn't define... how is the code doing that? If it's not doing anything special, why is this comment there? This is confusing.

Changed to:

```
     * Cancel bindings for performance, while ensuring to not cancel bindings
     * SegmentedBar didn't define by not doing `repetition.cancelBindings()`;
     * doing so will break Repetition after SegmentedBar enterDocument again. 
```

dharcourt: OK, now I understand. I didn't know the part about
repetition.cancelBindings(), as it hadn't occurred to me to call that.

These content/contentController bindings that are broken here, when will
they be restored? Does templateDidLoad() get called again when a component
is reused after an exitDocument? I suspect not.

This dynamic destruction of bindings is a code smell... why is it here?
We're trying to save the memory space for one binding when a SegmentedBar
is temporarily taken out of a document? Is that significant?

It seems better to just have the repetition.contentController binding in
the serialization and not mess with it.

No?

    - Charles

ui/segmented-bar.reel/segmented-bar.js
rayshan: > You can work around #1521 without introducing a dependency on a private variable by creating your own instance variable in SegmentedBar that tracks whether prepareForActivationEvents() was called. You can even take this opportunity to name the variable something that will be more understandable in this context, like wasSelectionBound or something. Having a comment referring to #1521 is a very good idea so the temporary variable can be removed in time.

> Using _preparedForActivationEvents will necessarily introduce a bug when that private variable is later renamed, so it's better to have #1521 cost the space for an extra boolean in every SelectedBar.

> Except in very extreme circumstances (which this is not) using private instance variables and private APIs is just plain laziness, often done by people who couldn't think of a better way to solve the problem they faced. You're better than that :-).

No longer need it now that we're setting bindings in setters :)

ui/segmented-bar.reel/segmented-bar.js
rayshan: > BUG: Line 211 will not dispatch a second action event when the user clicks on a segment, then clicks on that same segment again at a later time but before clicking on another segment. In general it is not possible to use a repetition selection binding to trigger a dispatchActionEvent(). Before you fix that bug you will probably want to write a test that tests for it.

Will fix w/ test.

rayshan: > BUG? Will the logic at line 250 work if the orientation is changed from horizontal to vertical, then back to horizontal? I haven't tested, but looking at this code it seems it won't work as expected...

Yes, missing behavior, will fix.

rayshan: > Declaring the _content, _contentController, and _orientation properties is unnecessarily verbose. Those private properties are only used in the content, contentController, and orientation setters and getters and they don't need to be declared outside of those. As an added bonus, keeping _content, _contentController, and _orientation to within their corresponding accessors will ensure those properties are created only if they are needed.

I see that everything works w/o those private variables. Why is it all over the place in the codebase? It seems to be an odd convention...

rayshan: > line 23: What? Why would an action event be tied to a selection flag? I might want to handle segmented bar clicks without being interested in the selection. Also, why is there an isSelectionEnabled flag in the first place? Are we trying to improve performance in those cases where the develop does not want to handle clicks and/or selection? How much time does this save? Is it worth adding an extra flag and concept that the developer wouldn't otherwise have to deal with?

Refactored out per discussion.

rayshan: > line 110: I think this is wrong... it seems that this will put in the event the "content" object of the selected segment, which is simply a cover for the segment's length and label properties when the segment is a Segment, and may be nothing or something entirely different when the segment is a custom class. What should be in the event is the Repetition:iterator.object corresponding to the segment clicked on.

Are you sure you're referring to the right line w/ the `@todo`? I'm confused.

rayshan: > Note that the whole selection binding set up in _defineSelectionBinding() is not necessary: The fact that a user has clicked on the segmented bar needs to be detected (which can't be done through a selection binding), and at that point the SegmentedBar can easily ask its repetition for selection0, so there's no need for a binding to get the appropriate data into the action event.

Refactored out per discussion.

rayshan: > BUG (I think): If you set a dataController value (legacy name: contentController :-), then let the segmented bar draw itself, then set a data value (legacy name: content), I don't think the segmented bar will correctly update the bindings to the repetition so the new data is displayed (because the bindings are set in templateDidLoad, which is called once and then never again). The most robust way to fix this is to ensure the segmented bar always has an RangeController, either taken or derived from the dataController or derived from the data. The repetition can then have a binding to this arrayController defined in the serialization and the binding doesn't need to change. Setters for data and dataController will ensure any changes get reflected in a correct _rangeController, which will update the repetition. Because you'll have better separation of concerns, the code for this might actually be smaller than the current version. No _data and _dataController properties will be necessary, and It's likely no needsDraw = 1 will be needed in the data and dataController setters because the binding will cause changes to _rangeController to propagate correctly to the repetition and force a redraw there. A test should be added for a dataController -> data change and vice versa.

Refactored out per discussion.

rayshan: > It doesn't seem right that SegmentedBar should have a childrenPath/childrenExpression that's redundant with TreeController's: If developers want to create a view on a tree that doesn't have the default structure, they can and should create a TreeController and set its childrenExpression.

Agreed, will fix

rayshan: > OK, now I understand. I didn't know the part about
repetition.cancelBindings(), as it hadn't occurred to me to call that.

> These content/contentController bindings that are broken here, when will
they be restored? Does templateDidLoad() get called again when a component
is reused after an exitDocument? I suspect not.

> This dynamic destruction of bindings is a code smell... why is it here?
We're trying to save the memory space for one binding when a SegmentedBar
is temporarily taken out of a document? Is that significant?

> It seems better to just have the repetition.contentController binding in
the serialization and not mess with it.

Will fix per discussion

rayshan:
All review comments addressed.

rayshan:
Per discussion, reinstated private property keys, e.g. `_data`, split Segment's binding into 2, `length` & `label`. All tests pass. Merging.

PR #1516, opened 2015-01-29
Accordion.reel component APIs review

kevinluosl:


rayshan:
If the `.info` is supposed to be a demo, I think we need to think through what should go in there. This commit seems to have an entire self-contained app in `.info`, with its own dependencies, assets, etc. I think a better approach could be having the demo and tests depend on the version of MontageJS it's committing against. This way we won't have 10 components with different dependencies in MontageJS, and any other change to MontageJS could be tested against whether they break any built-in components.

ui/accordion.reel/accordion.js
```diff
@@ -0,0 +1,152 @@
...
+ * @requires montage/ui/component
+ */
+var Montage = require("montage").Montage,
+    Component = require("montage/ui/component").Component;
+
+/**
+ * @class Accordion
+ * @extends Component
+ */
+exports.Accordion = Montage.create(Component, {
```
rayshan: Should we use `Montage.create(Component, {})` or `Component.specialize({})`? Other components seem to use the later. Not sure what the best practice is.

marchant: Montage.specialize(), Montage.create() is deprecated

ui/accordion.info/ui/accordion.reel/accordion.js
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * @function prepareForActivationEvents
+     * @param null
+     * @description Add the the event listener for this class object
+     * @private
+     */
+    prepareForActivationEvents: {
+        value: function () {
+            if (window.Touch) {
```
marchant: A press Composer needs to be used here, but with a repetition doing the work of repeating the array of components passed, the repetition handles the selection, so we might be able to get what we want from there

ui/accordion.info/ui/accordion.reel/accordion.js
```diff
@@ -0,0 +1,152 @@
...
+                this.element.addEventListener('touchend', this);
+            } else {
+                this.element.addEventListener('mouseup', this);
+            }
+        }
+    },
+
+    /**
+     * @function findClosestOfType
```
marchant: This should be completely unnecessary

ui/accordion.info/ui/accordion.reel/accordion.js
```diff
@@ -0,0 +1,152 @@
...
+     */
+    titles: {
+        value: null
+    },
+
+    /**
+     *@property activeIcon
+     *@description Object of icon for selected status
+     */
+    activeIcon: {
```
marchant: This is presentation related and doesn't belong in JavaScript API at all. We do not design widget that way in Montage

ui/accordion.info/ui/accordion.reel/accordion.js
```diff
@@ -0,0 +1,152 @@
...
+        value: null
+    },
+
+    /**
+     *@property normalIcon
+     *@description Object of icon for normal status
+     */
+    normalIcon: {
+        value: null
```
marchant: not in JS, see up

ui/accordion.info/ui/accordion.reel/accordion.js
```diff
@@ -0,0 +1,152 @@
...
+     */
+    expendAll: {
+        value: false
+    },
+
+    /**
+     * @property selectedOption
+     * @description Set the property to save the selected index of option when expendAll property is false.
+     */
+    selectedOption: {
```
marchant: Using the repetition, which handle the selection, we get this for free. If the accordion is going to be data driven, then it will have a rangeConroller, created internally from content(an array) or through contentController being set directly. We just had that discussion today with the segmentedBar

ui/accordion.info/ui/accordion.reel/accordion.js
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * @function handleEvent
+     * @param event -- event object
+     * @description Event handle function
+     * @private
+     */
+    handleEvent: {
+        value: function (event) {
+            event.preventDefault();
```
marchant: We shouldn't have to do all that

ui/accordion.info/ui/accordion.reel/accordion.js
```diff
@@ -0,0 +1,152 @@
...
+     */
+    headerBackgroundColor: {
+        value: null
+    },
+
+    /**
+     * @property title
+     * @description This is the label of header,it is an array.
+     */
+    titles: {
```
marchant: As I was working on the spec, it occurred to me that it would be much simpler to get the title/label/header from the component being displayed in each section of the accordion. So we wouldn't need that, but eventually a "titleExpression" which is how we would know what frb expresion to use to extract the display name from each component being displayed

ui/accordion.reel/accordion.js
kevinluosl: Yeah, I think so. Just need to know what's the correct way to add example, test and document in info folder, better to have a example

rayshan: @marchant thanks added to docs

PR #1520, opened 2015-02-05
Add Accordion component

kevinluosl:
This Accordion is implement after discuss with Charles and Javier, it includes 2 components, accordion.reel and a internal accordion-collapsible-panel.reel panel that contains header and body. When using accordion, people can pass in any components as header and body.

Accordion.content describes the data mode of passed in components

Main.reel is the example put in accordion.info/samples/ for now until we finally decide where to put the example codes.

Animation now is normal no-hardware accelerated height transition when switch among sections. Will try to use 3dTransition for this.

doc, test and meta is empty for now.


kevinluosl:
Re pull from

https://github.com/montagejs/montage/pull/1522

PR #1534, opened 2015-02-11
add .editorconfig per draft style guide

rayshan:
https://gist.github.com/rayshan/1ac3fd6329c03124e483


kriskowal:
Good idea. This should be applied throughout.

PR #1536, opened 2015-02-12
Collapsible panel

kevinluosl:
Collapsible panel third round review.

screen shot 2015-02-12 at 3 30 04 pm


ui/collapsible-panel.info/sample/main.reel/main.html
```diff
@@ -0,0 +1,271 @@
...
+            <div data-montage-id="example-4-repetition">
+                <div data-montage-id="example-4-collapsiblePanel" class="Main-CollapsiblePanel">
+                    <div data-montage-id="example-4-header" data-arg="header"></div>
+                    <div data-montage-id="example-4-content" data-arg="body"></div>
+                </div>
+            </div>
+        </div>
+    </div>
+</body>
+</html>
```
dharcourt: typo @ line 221: "ALorem" -> "Lorem"

dharcourt: spacing nit @ line 114: "[1,2]" -> "[1, 2]"
spacing nit @ line 175: "[1,3,4]" -> "[1, 3, 4]"

dharcourt: typo @ lines 137 & 245: "example-3-content1" -> "example-3-content-1"

dharcourt: spacing nit @ lines 67, 68, 69, 70, 71, 163, 164, 165, 166, 167: Space needed after each commas.

dharcourt: typo @ line 113 & 174: "false" -> "true"!

dharcourt: Clicking & dragging or double clicking in the header of each of these examples selects text. This behavior is strange, would never appear in a (good) shipping application, and should be fixed in this example.

ui/collapsible-panel.reel/collapsible-panel.css
```diff
@@ -0,0 +1,22 @@
...
+    -moz-transition: max-height 0.25s ease;
+    -o-transition: max-height 0.25s ease;
+    transition: max-height 0.25s ease;
+}
+
+.montage-CollapsiblePanel--active {
+    height: auto;
+    /* A popular way to animation auto height, we have to set a max-height to trigger the animation */
+    max-height: 1000px;
+}
```
dharcourt: lines 12, 13, & 14: Are the -webkit, -moz, and -o prefixes necessary? http://caniuse.com/#feat=css-transitions suggests no current version of Chrome, Safari, Firefox, or Opera requires these, and we're not supporting legacy versions of those browsers.

dharcourt: nit @ line 20: Period needed at the end of the sentence.

ui/collapsible-panel.info/sample/main.reel/main.html
kevinluosl: All fixed. Disabled text-select in header. We need prefix -webkit-, -ms-, and -moz- for latest version of these browser

dharcourt: line 17: "isMultipleSelectionAllowed" -> "isMultipleChoiceAllowed"

ui/collapsible-panel.reel/collapsible-panel.js
```diff
@@ -0,0 +1,126 @@
...
+            if (firstTime) {
+                header = this.element.children[0];
+                this._pressComposer = new PressComposer();
+                // Register composer here and add event in prepareForActivationEvents.
+                this.addComposerForElement(this._pressComposer, header);
+            }
+        }
+    }
+
+});
```
dharcourt: line 51: "(value instanceof ChoiceController)" seems dangerous... Why do you need to check if it's a ChoiceController? What if it was another object the developer wanted to provide that worked like a ChoiceController? Do we really need to check types here?

dharcourt: line 54: Can there be several bindings to a component property? What would happen if the developer sets up a separate binding to isExpanded for their own purposes? Wouldn't this code potentially break their binding?

kevinluosl: I check ChoiceController because I think we may only accept ChoiceController here (or extend class of ChoiceController). If we don't need to be this only Class, I think we can not check the Type.

dharcourt: I could be wrong (Javier/Benoit would probably know) but don't think instanceof checks are a common pattern in Montage. I think class checking is actually an anti-pattern that comes from static typed languages like Java. In JavaScript duck typing seems like the better way to go, and instanceof checks can be avoided.

kevinluosl: Regarding the binding. If we do binding like this way. User can't bind other stuff to isExpanded (except they bind after expansionController I think). What happen fi we don't check to cancel binding? So user can bind their own?

dharcourt: Can't you simply bind to _isExpanded? In any case, we need one way or the other: 1) to allow users to bind to isExpanded; and 2) to clean up bindings to previous ChoiceControllers if expansionController is changed.

kevinluosl: @dharcourt I removed instanceof checking for ChoiceController type so we allow people to pass in their own Controller. I checked this._expansionController.hasOwnProperty('choice') && this._expansionController.choice.hasOwnProperty('data') to make sure isExpanded binding can work correctly.  I think we can't allow user to binding isExpand outside because we bind it when setting expansionController and another binding will cause duplicate binding exception. In my opinion, we may do one way or another, 1) isExpanded is binding when setting _expansionController and we allow people to do their own _expansionController like our current way. I changed isExpanded to private too, so only use expansionController. 2) we binding isExpanded outside like the old way, don't do _expansionController but directly bind isExpanded outside so user can more free when using isExpanded

PS: I checked that montage uses lots instancef in other classes so I guess it's a correct way to use if we really want to check a instance Class type.

ui/collapsible-panel.info/sample/main.reel/main.html
```diff
@@ -0,0 +1,291 @@
...
```
dharcourt: What about removing `example1Choice`, `example2Choice`, `example3Choice`, and `example4Choice` by replacing syntax like `"choice": {"@":"example1Choice"}` with the `"choice": {"data": [2]}` syntax we discussed on Skype?

kevinluosl: @dharcourt Yeah, update for this change.

ui/collapsible-panel.reel/collapsible-panel.js
dharcourt: Ouch... the hasOwnProperty() is more ugly than was the instanceof it replaces. I don't understand why any of these checks are necessary... What situation are you trying to prevent? Are you trying to make sure your code will work with all possible passed in values for expansionController? That's not possible: In a dynamic language like JavaScript no amount of checking will "make sure isExpanded binding can work correctly". Are you trying to catch errors developers are likely to make? How likely are they to make the mistake you're checking for, and in what way is the error message you're generating better than the error message they would have gotten without the extra type checking code?

My advice is: Remove all type checking and let the errors that naturally would occur if incorrect values are set occur. This will make the code simpler and less brittle.

dharcourt: I think this version is worse than the previous one. Either I'm not understanding what you're trying to do or the limitations you have or you're not understanding why I made the suggestions I did. In either case we should discuss offline to resolve the misunderstanding or this code review will be a waste of both of our time.

kevinluosl:
Rebase with master to fix conflict in run.js

hthetiot:
@kevinluosl Can you rebase on #master

PR #1545, opened 2015-02-19
fix "'Range.detach' is now a no-op" warning

### LINKED ISSUE #1523: "'Range.detach' is now a no-op" warning
It's deprecated per DOM specs (http://dom.spec.whatwg.org/#dom-range-detach), never actually did anything useful, most of other libraries that use it simply just removes it.

Remove:
https://github.com/montagejs/montage/blob/master/ui/loader.reel/loader.js#L494

rayshan:
closes #1523

### DIFF
--- a/ui/loader.reel/loader.js
+++ b/ui/loader.reel/loader.js
@@ -491,7 +491,6 @@ exports.Loader = Component.specialize( /** @lends Loader# */ {
 
                 if(this.removeContentOnLoad || this._forceContentRemoval) {
                     this._contentToRemove.extractContents();
-                    this._contentToRemove.detach();
                     this._contentToRemove = null;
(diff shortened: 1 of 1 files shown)


michaelshan1955:
I lost the contact with you for years since last time you called from HK, at that time i was still working for Lenovo. Now I have retired 2 years before official retiretime due to the company would like to pay me for the next two years until my retire age. Miss you very much, tried very hard to contact you but failed all the time, now at least I now you are in CA and working as an engineer like me. Wish you could contact me at my emailL 13714625960@163.com. wish you all the best!

michaelshan1955:
My email is my cell phone number + @163.com, so you will also know my cell phone.

PR #1548, opened 2015-02-24
Remove q-io dep

rayshan:
Done per @marchant 

I `promiseify`'ed `fs` with bluebird instead of using `fs` callbacks to not break the existing control flow. I have no idea what `node.js` module's usage is. There are no docs, nor tests. I replaced `q-io` methods based on its documentations and the most similar `fs` / `path` equivalent. 

All other tests pass (besides the 18 that have been failing).

Also fixed all the whitespace issues.

### DIFF
--- a/node.js
+++ b/node.js
@@ -1,12 +1,13 @@
 /*jshint node:true, browser:false */
 
-var FS = require("q-io/fs");
+var Promise = require("bluebird");
+var fs = Promise.promisifyAll(require("fs"));
+var path = require("path");
 
 var MontageBoot = require("./montage");
 
 var Require = require("mr/require");
 require("mr/node");
-var Promise = require("q");
 var URL = require("url");
 
 var htmlparser = require("htmlparser2");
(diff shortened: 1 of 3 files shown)


rayshan:
> Try running mop, that might trigger that code - @marchant

Did, mop tests pass, mopping results seem to be correct, I don't think it's invoking this files though.

rayshan:
Hi @kriskowal / @Stuk, would you mind letting us know what `node.js` is for? I can comment the file with your notes.

Stuk:
That file is for running parts of Montage in Node, and yes it is used by Mop. However IIRC Mop (arguably incorrectly) uses the version in its node_modules directory, instead of the one in the project you are mopping. @kriskowal can shed more light.

rayshan:
Thanks @Stuk, I did run mop on a project with updated `node.js` in `node_modules/montage`, not on this repo itself. Everything seem to be fine. Maybe eventually we need to move this file to mop. For now looks like we're ok.

PR #1557, opened 2015-03-20
Add Succession, Transition & enable Component build-in / build-out CSS animation

rayshan:
WIP, API not final, do not merge
===

`Component` is enhanced to allow animation for both stand-alone components & components inside a placeholder like `Slot` or `Substitution`.

`Succession` implements a `stack` to keep track of navigation. You can `push` / `pop` / `clear` the stack. You can also bind to the `top` / `previous` / `first` components of the stack, e.g. to a Navigation Bar. The stack only tracks Components for now. Next step is to enable the use of a `Transition` object.

See `succession.info` for usage. For live demo, serve `/index.html` then select `Succession` from drop-down.

Video demo: https://www.youtube.com/watch?v=9F0SxMVXpgI

API & Usage Docs: https://gist.github.com/rayshan/fd5ef874a3c648b81138

**Known issues**

- When pushing onto then popping components off of the stack rapidly, sometimes components get stuck on the stack. It's likely the way `_performDomContentChanges` removes children elements need improvement.
- ~~Firefox isn't consistently hearing transition / animation end events~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,9 +35,11 @@
     "q-io": "~1.11.0"
   },
   "devDependencies": {
+    "animate.css": "^3.1.1",
     "jshint": "~2.4.4",
     "montage-testing": "~0.4.1",
-    "native": "~0.2.0"
+    "native": "~0.2.0",
+    "normalize.css": "^3.x"
   },
   "scripts": {
     "test": "node test/node/node-spec.js",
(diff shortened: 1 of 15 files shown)


rayshan:
Closing in favor of https://github.com/montagejs/montage/pull/1559

PR #1559, opened 2015-03-28
Add Succession, Transition & enable Component build-in / build-out CSS animation

rayshan:


ui/component.js
```diff
@@ -3540,6 +3944,28 @@ var RootComponent = Component.specialize( /** @lends RootComponent.prototype # *
...
+     * Determine end of animation event name across browsers.
+     * Mainly for Chrome <= 41 & Safari <= 8.
+     *
+     * @readonly
+     * @property {String}
+     */
+    CSSAnimationEndEventName: {
+        get: function () {
+            if (!this._CSSAnimationEndEventName) {
+                if (window.onanimationend === undefined && window.onwebkitanimationend !== undefined) {
```
thibaultzanini: not safe use void 0 or typeof

thibaultzanini:
@rayshan Where are we about this PR ?

rayshan:
This is fully working and we're ready to use my fork with it for Contour (pending other Montage bugs being fixed). However it's not mergable due to missing tests.

thibaultzanini:
Ok thanks. what issues ?

rayshan:
All known issues are fixed. We're using it now in Contour without any issues. The missing feature is animating an entire subtree of components upon removal, instead of animating just the topmost component. Also missing tests.

rayshan:
@marchant debug steps for `getComputedStyle` property in `willDraw` breaking build-in transition:

- Serve project's index.html
- Select Succession in top left select control
- Scroll down to pink component called `baz`, which is wrapped by a condition and controlled by the toggle button; the code that causes the bug is in https://github.com/rayshan/montage/blob/feature-succession/ui/succession.info/sample/ui/baz.reel/baz.js
- To change from CSS transition to CSS animation, which isn't broken, you can swap between these: https://github.com/rayshan/montage/blob/feature-succession/ui/succession.info/sample/ui/baz.reel/baz.js#L8-L15
- To see that it's accessing the properties that cause the bug, not calling `getComputedStyle` itself, you can comment out https://github.com/rayshan/montage/blob/feature-succession/ui/succession.info/sample/ui/baz.reel/baz.js#L24-L25
- You can move the `willDraw` code to `draw`, which works

Let me know if I can offer further assistance.

PR #1560, opened 2015-04-08
Fix areStylesLoaded in document-resources

romancortes:


marchant:
Added some comments, please take a look!

marchant:
LGTM

PR #1565, opened 2015-05-14
Fixed RangeController issues

rayshan:
Discovered during work on Succession
- Ensure selection always has content
- Use swap instead of clear / add to prevent setting selection
  multiple times

### DIFF
--- a/core/range-controller.js
+++ b/core/range-controller.js
@@ -46,7 +46,7 @@ var _RangeSelection = function(content, rangeController) {
     //Moved to RangeSelection.prototype for optimization
     //self.makeObservable();
     self.__proto__ = _RangeSelection.prototype;
-    
+
     self.rangeController = rangeController;
     self.contentEquals = content && content.contentEquals || Object.is;
 
@@ -94,11 +94,11 @@ Object.defineProperty(_RangeSelection.prototype, "push", {
           while (++i < l) {
             x[i] = arguments[i];
           }
-        
+
         this.swap_or_push(this.length, 0, x);
     }
 });
-    
+
 /**
  * A custom version of swap to ensure that changes obey the RangeController
  * invariants:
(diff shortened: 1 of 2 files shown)


rayshan:
Reviewed by Javier, updated `handleAdd` logic

thibaultzanini:
@rayshan How can I reproduce it ?

rayshan:
You can run the updated test without the fix (cherry pick the spec file). You'll see that selection is set to null, then set to a new value, in 2 steps.

core/range-controller.js
```diff
@@ -773,7 +773,13 @@ var RangeController = exports.RangeController = Montage.specialize( /** @lends R
                 // remove all values from the selection that were removed (but
                 // not added back)
                 minus.deleteEach(plus, equals);
-                if (this.selection) {
+
+                if (this.selection.length) {
+                    // ensure selection always has content
+                    if (this.content.length && this.avoidsEmptySelection && !this.multiSelect) {
```
thibaultzanini: If content is equal to null that should fail

core/range-controller.js
```diff
@@ -773,7 +773,13 @@ var RangeController = exports.RangeController = Montage.specialize( /** @lends R
...
+                        // selection can't contain previous content value as content already changed
+                        this.selection.add(this.content[this.content.length - 1]);
+                    }
```
thibaultzanini: why do we select the last element ?

core/range-controller.js
rayshan: This is nested inside another length check:
https://github.com/rayshan/montage/blob/pr-rangecontroller/core/range-controller.js#L771
...which should probably check for `this.selection` existence first, but I didn't want to touch code that wasn't considered broken. There are quite a few other `length` checks in this file that can be rewritten.

core/range-controller.js
rayshan: Because user wants `avoidsEmptySelection`. I discussed with @marchant and we concluded the right behavior is to select the previous item in the array.

core/range-controller.js
thibaultzanini: Ok I got it. You assume that whether the selection is not empty technically the content should not be empty as well. Looks like we have some security for the selection property: https://github.com/rayshan/montage/blob/pr-rangecontroller/core/range-controller.js#L468

core/range-controller.js
thibaultzanini: @marchant do you have an real user case as an exemple ?

rayshan: An example would be a table view bound to data fetched from a server. If the user clicks on a table cell to select it, then server deletes that cell's data, which causes the cell to be removed, what should selection include? `avoidsEmptySelection` should guarantee that selection will never be empty, because there may be code depending on its content.

thibaultzanini: Ya I got that, but why the last one ? it seems just funny to me but it's a personal thought.

thibaultzanini: I would maybe prefer something close to the last visible index or maybe leave also the choice to the developper with a delegate method? @romancortes any thoughts?

thibaultzanini:
Have already been merged!
https://github.com/montagejs/montage/commit/fb426e0e8274450af2c0d84f73a7f1e4163f957a

PR #1575, opened 2015-06-16
Fix prepareForDraw error and related

### LINKED ISSUE #1574: prepareForDraw causes a warning and an error
Defining prepareForDraw in any child of Component causes a warning: "Montage.callDeprecatedFunction is deprecated" and an error: "Uncaught ReferenceError: Montage is not defined", resulting in the page not loading

cdebost:
Fixes #1574


hthetiot:
Not valid with master.

PR #1576, opened 2015-06-19
Quick fix when trying to require a JSON file within a mopped app.

thibaultzanini:
:warning: do not merge. WIP

PR probably outdated, need to fix mr instead.

Within a mopped app the JSON content is already parsed and the property `text` is undefined.

### DIFF
--- a/packages/mr/require.js
+++ b/packages/mr/require.js
@@ -782,8 +782,7 @@
 
     Require.JsonCompiler = function (config, compile) {
         return function (module) {
-            var json = (module.location || "").match(/\.json$/);
-            if (json) {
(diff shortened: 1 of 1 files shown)


rayshan:
Can we merge this please? It works. Not sure why WIP.

thibaultzanini:
I would need to investigate a little bit more and I would probably need to discuss about it with @marchant.

rayshan:
ok thx

thibaultzanini:
PR probably outdated, need to fix mr instead.

rayshan:
I need to keep an eye on mr as this patch is used in montage-contour fork and need to be backed out after mr is patched.

PR #1579, opened 2015-07-01
Fix super direct infinite loop in observed getter/setters

romancortes:


dharcourt:
Looks good to me!

marchant:
LGTM

PR #1581, opened 2015-07-02
Enhance Repetition performance and fix bug in selection

romancortes:
Enhances Repetition performance by replacing an expensive binding for the selected property in the repetition's iterations by a getter/setter and the additional logic to maintain it up to date.

Refractors a couple of pieces of code by removing an unnecessary loop and merging 2 loops into one.

Fixes a bug in selection when used in conjunction with visibleIndexes from the repetition's controller and adds a test for this bug.

### DIFF
--- a/ui/repetition.reel/repetition.js
+++ b/ui/repetition.reel/repetition.js
@@ -755,21 +755,33 @@ var Repetition = exports.Repetition = Component.specialize(/** @lends Repetition
     handleSelectionRangeChange: {
         value: function (add, remove) {
             var iterationsMap = new Map(),
-                iteration,
                 length = this.iterations.length,
-                i;
+                objectIterations,
+                object,
+                iteration,
+                i, j;
 
             for (i = 0; i < length; i++) {
-                iterationsMap.set(this.iterations[i].object, this.iterations[i]);
+                iteration = this.iterations[i];
+                object = iteration.object;
+                if (!(objectIterations = iterationsMap.get(object))) {
+                    objectIterations = [];
+                    iterationsMap.set(object, objectIterations);
+                }
+                objectIterations.push(iteration);
             }
             for (i = 0; i < remove.length; i++) {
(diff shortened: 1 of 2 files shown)


ui/repetition.reel/repetition.js
```diff
@@ -700,6 +719,62 @@ var Repetition = exports.Repetition = Component.specialize(/** @lends Repetition
...
+                    );
+                    this.handleSelectionRangeChange(value, []);
+                } else {
+                    this._cancelSelectionRangeChangeListener = null;
+                }
+            }
+        }
+    },
+
+    handleSelectionRangeChange: {
```
marchant: Is it really worth creating a new Map (garbage) every time the selection changes, to find the iteration object, compared to doing a loop ourselves on iterations to find this.iterations[i].object === aObjet in add/remove?

romancortes: In a multi-selection scenario the loop would end being in worst case O(n ^ 2). Suppose we have a million iterations and we perform a select all function... that would be very slow.

So I'm using the Map as a hash, and I assume the hash look up might range depending on the implementation between O(1) and O(log n), so in the worst case scenario - select all - I would end with a maximum of O(n log n).

As the most common case for selection is not multi-selection, what I propose is to add a conditional in the code for simple selection, if add and remove lengths are lower or equal to 1, then I could just loop and then I would avoid the Map creation in that case. Does it sound good?

marchant: That makes sense, to loop for small sizes, but maybe also create a shared map that can be emptied and reused?

romancortes:
I found a bug, so I'm closing temporarily this PR. The bug was assuming the iterations and iteration.objects are bijective, but that is not true; multiple iterations can point to the same object. I'll work again on this and then do a new PR.

romancortes:
Fixed bugs, PR open again

PR #1585, opened 2015-07-24
Initial commit for the tap-composer.

thibaultzanini:
:warning: devices with multiple pointers are not supported yet.


thibaultzanini:
@marchant PR Updated.

PR #1586, opened 2015-07-25
Event-Manager needs to listen on mouseDown or touchStart events for devices with multiple pointers.

thibaultzanini:
Event-Manager needs to listen on both mouseDown or touchStart events for devices with multiple pointers in order to activate components. Plus, this commit fixes a issue for Firefox that implement by default the Touch constructor even if it’s not supported by the device.

[EDIT] Do not merge yet


thibaultzanini:
@marchant, I'm not sure but it's probably in order to set a component as active when its element can take the focus when you hit tab for exemple.

PR #1587, opened 2015-07-27
Fix needsClearVisibleIndexes in Flow

romancortes:
Fixes bug preventing selection when clicking for the first time on an
iteration.

### DIFF
--- a/ui/flow.reel/flow.js
+++ b/ui/flow.reel/flow.js
@@ -1290,6 +1290,11 @@ var Flow = exports.Flow = Component.specialize( /** @lends Flow.prototype # */ {
                 j,
                 i;
 
+            if (this._needsClearVisibleIndexes) {
+                this._visibleIndexes.splice(newVisibleIndexes.length, Infinity);
+                this._needsClearVisibleIndexes = false;
+            }
+
             // Search for viable holes, leave content at the same visible index
             // whenever possible.
             for (i = 0; i < oldIndexesLength; i++) {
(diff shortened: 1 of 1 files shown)


ui/flow.reel/flow.js
```diff
@@ -1290,6 +1290,11 @@ var Flow = exports.Flow = Component.specialize( /** @lends Flow.prototype # */ {
                 j,
                 i;
 
+            if (this._needsClearVisibleIndexes) {
+                this._visibleIndexes.splice(newVisibleIndexes.length, Infinity);
```
marchant: Are you sure about this splice(), the use of Infinity vs doing this._visibleIndexes.length - newVisibleIndexes.length ?

romancortes:
From http://www.ecma-international.org/ecma-262/5.1/ :

15.4.4.12 Array.prototype.splice (start, deleteCount [ , item1 [ , item2 [ , … ] ] ] )
...
Let actualDeleteCount be min(max(ToInteger(deleteCount),0), len – actualStart).
...
According to the spec, ToInteger(Infinity) will result in Infinity. Max of Infinity and 0 is Infinity. Min of Infinity and len - the length of the array - minus actualStart is len - actualStart, that is exactly what we are searching for. 

I have tested in all latest major browsers - Opera, Safari, Firefox, IE, Chrome - and they follow the spec. In general, Infinity is very well handled by Javascript.

marchant:
Ok!

PR #1589, opened 2015-07-28
Add setter for AbstractTextField.hasFocus

BnTejaede:
Allow bindings to the public hasFocus trigger on property change by using a setter rather than setting the private _hasFocus directly.


rayshan:
Cherry picked into https://github.com/montagestudio/montage-contour#feature-succession

PR #1593, opened 2015-08-18
Changes how the repetition select its iteration by using the PressComposer and improvements

thibaultzanini:
Cancel selection when a scroll starts.
Edit: :warning: needs this PR to be merged first https://github.com/montagejs/montage/pull/1594

### DIFF
--- a/ui/repetition.reel/repetition.js
+++ b/ui/repetition.reel/repetition.js
@@ -7,6 +7,7 @@ var Template = require("../../core/template").Template;
 var RangeController = require("../../core/range-controller").RangeController;
 var Promise = require("../../core/promise").Promise;
 var browser = require("../../core/browser").browser;
+var PressComposer = require("../../composer/press-composer").PressComposer;
 
 var Map = require("collections/map");
 var Set = require("collections/set");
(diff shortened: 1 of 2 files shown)


ui/repetition.reel/repetition.js
```diff
@@ -1881,57 +1893,45 @@ var Repetition = exports.Repetition = Component.specialize(/** @lends Repetition
...
-            } else {
-                // dispatches handleMouseup
-                document.addEventListener("mouseup", this, false);
-                // dispatches handleMousemove
-                document.addEventListener("mousemove", this, false);
+                this.__pressComposer.addEventListener("press", this, false);
+                this.__pressComposer.addEventListener("pressCancel", this, false);
+
+                this.element.addEventListener("touchmove", this, false);
+                document.addEventListener("scroll", this, true);
```
marchant: Why isn't mouse move here as well? Is that really needed? Why isn't it abstracted through composer?

ui/repetition.reel/repetition.js
```diff
@@ -1942,94 +1942,32 @@ var Repetition = exports.Repetition = Component.specialize(/** @lends Repetition
...
-                document.removeEventListener("touchmove", this, false);
-                document.removeEventListener("touchcancel", this, false);
+            this.__pressComposer.removeEventListener("press", this, false);
+            this.__pressComposer.removeEventListener("pressCancel", this, false);
 
-            } else {
-                document.removeEventListener("mouseup", this, false);
-                document.removeEventListener("mousemove", this, false);
-            }
+            this.element.removeEventListener("touchmove", this, false);
```
marchant: Symmetric comment, why do we still need this here even though we've abstracted most of the work through a composer?

ui/repetition.reel/repetition.js
thibaultzanini: We already talked about it and we came up with this solution to disable an iteration when we are scrolling. Listening to mousemove is not needed here because the PressComposer will check if we are still on the same target when we release the mouse. Plus, I believe we can't scroll with mousemove, right ?

It's not managed by the PressComposer because it can be tricky in some cases, for example with a big button we can still move on it between a TouchStart and a TouchEnd.

thibaultzanini:
PR updated

thibaultzanini:
PR updated

marchant:
Reviewed with @thibaultzanini Gtg

PR #1594, opened 2015-08-21
Add the property eventPhase in order to specify on which phase of the  event flow composers should work.

thibaultzanini:
Feature only supported by the PressComposer at the moment.

### DIFF
--- a/composer/composer.js
+++ b/composer/composer.js
@@ -45,6 +45,50 @@ exports.Composer = Target.specialize( /** @lends Composer# */ {
         }
     },
 
+    /**
+     * Specifies on which phase of the event flow composers should work.
+     *
+     * @property {boolean}
+     * @public
+     * @default true - Capture Phase
+     */
+    eventPhase: {
+        get: function () {
+            return this._eventPhase;
+        },
+        set: function (eventPhase) {
+            var newEventPhase;
+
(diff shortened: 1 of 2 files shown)


marchant:
Looks good, only question is what happens if eventPhase is changed after load(), when eventListeners are in place already? It should probably either throw or unload/load again in the the otherEventPhase

thibaultzanini:
Agree I'll the PR

composer/composer.js
```diff
@@ -45,6 +45,50 @@ exports.Composer = Target.specialize( /** @lends Composer# */ {
...
+                newEventPhase = false;
+            }
+
+            if (typeof newEventPhase !== "undefined" && newEventPhase !== this._eventPhase) {
+                if (this._isLoaded) {
+                    this._unload();
+                    this._eventPhase = newEventPhase;
+                    this._load();
+
+                    if (typeof this._handleEventPhaseChange === "function") {
```
thibaultzanini: @benoitmarchant using a flag named for example `eventPhaseChanged` is not good enough here. 
It worked well when we are in the middle of the press for example, we can clean the composer (remove/add listeners) at the end of the interaction but where the composer is supposed to clean itself if we change the event phase before an interaction?

PR #1599, opened 2015-08-26
Cancel Press Event if the touch end target is not the press composer element or one of its children

thibaultzanini:


marchant:
LGTM!

PR #1602, opened 2015-09-01
Add the property isLoaded to composers.

thibaultzanini:


ui/overlay.reel/overlay.js
```diff
@@ -171,6 +154,8 @@ exports.Overlay = Component.specialize( /** @lends Overlay.prototype # */ {
 
                 _window = this.element.ownerDocument.defaultView;
                 _window.addEventListener("resize", this);
+
+                this._pressComposer = new PressComposer();
```
marchant: This should be in prepareForActivationEvent, not in enterDocument

thibaultzanini: @benoitmarchant I don't think so, given that it's an overlay we need to be able to know if an user click outside an overlay in order to hide that overlay.

thibaultzanini: We could instead create it when this overlay is shown for the first time.

ui/component.js
```diff
@@ -2385,15 +2385,44 @@ var Component = exports.Component = Target.specialize( /** @lends Component.prot
...
     },
 
     /**
+     * Load a Composer
+     * @function
+     * @param {Composer} composer
+     */
+    loadComposer: {
+        value: function (composer) {
+            if (this.composerList.indexOf(composer) > -1) { // [Need discussion] -> composer.component === this
```
thibaultzanini: FYI @marchant need discussion.

PR #1604, opened 2015-09-11
Fixes the Text Component in order to display some falsy values.

thibaultzanini:
The Text Component wasn't able to display the string representation of the boolean false.


rayshan:
Should we stringify the value per here? https://github.com/montagejs/montage/pull/1204#issuecomment-75109349

PR #1607, opened 2015-09-18
Reduces memory consumption within Component.

thibaultzanini:
Not creating by default arrays for the properties `composerList` and `composerListSwap`.


marchant:
LGTM

PR #1597, opened 2015-08-24
Adds support for devices with multiple pointer types to the Translate Composer and improvements.

thibaultzanini:
- Support Pointer Events
- Support devices with multiple pointer types.
- Performance improvements. (less translateStart raised when handling wheel event, less listeners and diverse improvements)
- Better stealing claimed pointers logic, wait for a reel move before stealing the pointer. The logic is based on different mouse and touch radius, except between Translate Composers the fact of stealing is immediate between these Composers.
- Fixes issues between composers that was claiming the same pointer. -> For example, using the translate composer on an element within a repetition. (repetition uses the PressComposer).
- Doesn’t listener on wheel events by default. (feature just used by the flow and scrollers)
- Fixes bug between inner scroller Components. (Was not able to work)
- Add unload function. (release listeners)
- Updates translate composer tests.
- Backwards compatibility for the Flow Translate Composer.
- Updates Event Manager in order to support storage of Pointer Events.
- Adds supports for horizontal “wheel” movement.
- Stops to raise translate events when a movement is beyond a extreme (max/min Translate X/Y).

PR #1612, opened 2015-09-29
Makes Montage works with PointerEvents

thibaultzanini:
:warning: needs this PR to be merged first https://github.com/montagejs/montage/pull/1597

- Adds an option to disable the filtering of emulated mouse events.
- Disables the filtering of emulated mouse events for browsers that support Pointer Events.
- Fixes an issue that was blocking the logic for prepareForActionEvents, indeed ”ontouchstart" exists even on non-touch devices under Firefox.
- Fixes tests.

### DIFF
--- a/core/event/event-manager.js
+++ b/core/event/event-manager.js
@@ -2283,43 +2283,48 @@ if (typeof window !== "undefined") { // client-side
             }
         },
 
+        blocksEmulatedEvents: {
+            value: true
+        },
+
         /**
          * @function
          * @param {Event} event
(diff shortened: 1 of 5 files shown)


core/event/event-manager.js
```diff
@@ -2265,6 +2265,10 @@ if (typeof window !== "undefined") { // client-side
             }
         },
 
+        isEmulatedEventFiltered: {
```
thibaultzanini: blocksEmulatedEvents

PR #1613, opened 2015-09-30
Composers should cancel their work when in some new cases.

thibaultzanini:
- The TranslateComposer cancel a translate when a scroll event is raised.
- The PressComposer cancel a press when its element’s position has changed or when a scroll event is raised.
- The repetition doesn’t handle scroll, move anymore.


marchant:
Reviewed, LGTM!

PR #1615, opened 2015-10-02
PrepareForActivationEvents should be called when a pointing device (mouse/touch) is moved over a component’s element.

thibaultzanini:


marchant:
Reviewed

PR #1617, opened 2015-10-02
Move visibleIndexes from RangeController to Repetition

romancortes:
This change slightly enhances performance on all repetitions and increases considerably performance for Flow.


ui/flow.reel/flow.js
```diff
@@ -1615,34 +1611,34 @@ var Flow = exports.Flow = Component.specialize( /** @lends Flow.prototype # */ {
...
-                    offset = this.offset(visibleIndexes[i]);
+                    offset = this.offset(this._visibleIndexes[i]);
                     pathIndex = offset.pathIndex;
                     slideTime = offset.slideTime;
                     indexTime = this._splinePaths[pathIndex]._convertSplineTimeToBezierIndexTime(slideTime);
                     iteration = this._repetition._drawnIterations[i];
                     element = iteration.cachedFirstElement || iteration.firstElement;
                     if (indexTime !== null) {
-                        if (element.children[0]) {
+                        if (elementChildren = element.children[0]) {
```
marchant: elementChildren is cached but it's only used for elementChildren.classList, why not cache elementChildrenClassList?

romancortes: Because there is a conditional there. First I check if there is an element.children[0], then I can access the classList, else if I try to do directly element.children[0].classList I would get an error because children[0] might be undefined

romancortes: On the other hand, the correct name would probably be elementFirstChild

thibaultzanini:
Todo-mvc: Failed to load.
Popcorn: Detail panel is empty at the first load.

marchant:
Tested in apps

PR #1619, opened 2015-10-05
Remove an unnecessary closure since the function `_handleWheelTimeout` is bound to the translate composer.

thibaultzanini:


marchant:
LGTM

PR #1620, opened 2015-10-08
Remove unnecessary super calls in Component & Target for performance

rayshan:
Per discussion with @marchant. No new failing tests. Tested in Contour with noticeable improvement.


thibaultzanini:
LGTM!

PR #1624, opened 2015-10-16
Load speedup

marchant:


thibaultzanini:
Looks like we earn about 15ms with the minit default app. :)

thibaultzanini:
Tests pass. LGTM.

PR #1629, opened 2015-10-23
Makes sure a component detaches all its child components when its domContent change.

thibaultzanini:
The property childComponents is an array, so the length is changing when an element is removed of it. Therefore, all child components were not detached.


thibaultzanini:
@benoitmarchant :warning: Critical bug!!

marchant:
Good catch!

PR #1637, opened 2015-11-08
Remove references to "lifecyle" from documentation

dharcourt:
There are a couple of places in the documentation that have unnecessary
references to the concept of an object life cycle. This commit removes
them to simplify the documentation.

There are also a couple of places in the code where comments refer to
"lifecycle". This commit does not attempt to get rid of these but it
does correct the misspelling in the comments to "life cycle".


thibaultzanini:
@dharcourt there is some conflict.

hthetiot:
lgtm

PR #1640, opened 2015-11-13
Better support cordova

thibaultzanini:
- Add support of events filtering for UIWEbViews.
- Checks if the element position of the PressComposer has changed during a scroll event.


thibaultzanini:
FYI @benoitmarchant @rayshan

PR #1643, opened 2015-11-19
Fix Loader indicators.

thibaultzanini:
Makes sure the minimum amount of time for the bootstrapping and the loading indicator is respected. Updates the Loader’s specs.


marchant:
Reviewed

PR #1646, opened 2015-11-30
Minor performance improvements.

thibaultzanini:
- Use brackets to get characters instead of using the `charAt()` method.
- Use the `charCodeAt` method instead of the `charAt()` method, up to 40% faster under Chrome 46 and 30% under Safari 9, no noticeable difference under Firefox.


marchant:
LGTM

PR #1651, opened 2015-12-28
Event manager improvements

marchant:
Removes “use strict” in meta/* as it impacts super() implementation on Safari 9.0.2 (11601.3.9), 	Removes dependency on uuid in EventManager, optimizes dispatching logic


thibaultzanini:
LGTM

PR #1656, opened 2015-12-30
Make the loadScripts method faster.

thibaultzanini:


marchant:
LG, avoids unnecessary array creation

PR #1657, opened 2015-12-30
Eliminate creation of a new Promise in Component

rayshan:
Had this one sitting in my stash.

### DIFF
--- a/ui/component.js
+++ b/ui/component.js
@@ -1356,15 +1356,11 @@ var Component = exports.Component = Target.specialize( /** @lends Component.prot
             if (!this._expandComponentPromise) {
                     if (this.hasTemplate) {
                         var self = this;
-                        this._expandComponentPromise = new Promise(function(resolve, reject) {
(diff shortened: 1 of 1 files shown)


thibaultzanini:
LGTM

francoisfrisch:
This change is not equivalent, it means that the promise will not reject if
instantiateTemplate fails for some reason. You should throw in the catch.

thibaultzanini:
Thanks for the feedback @francoisfrisch. :)

I think I'm going to remove the catch statement anyway.

It's kind of redundant, because errors are already "tracked" within the `_instantiateTemplate` function:
- https://github.com/montagejs/montage/blob/master/ui/component.js#L1465
- https://github.com/montagejs/montage/blob/master/ui/component.js#L1486 

Plus, calling the error function of the console object in the catch gives you the following error: `TypeError: Illegal Invocation`.

rayshan:
Thanks guys.

PR #1658, opened 2015-12-31
Avoid to create unnecessary Promise and fix wrong indentation.

thibaultzanini:


ui/component.js
```diff
@@ -1475,13 +1471,13 @@ var Component = exports.Component = Target.specialize( /** @lends Component.prot
...
-                return template.instantiateWithInstances(instances, _document)
-                .then(function(documentPart) {
+                return template.instantiateWithInstances(instances, _document).then(function(documentPart) {
                     documentPart.parentDocumentPart = self._ownerDocumentPart;
                     self._templateDocumentPart = documentPart;
                     documentPart.fragment = null;
                     instances = null;
-                },function (reason) {
+
+                }, function (reason) {
```
rayshan: I think it's customary to use `error` as argument name due to the type.

rayshan:
LGTM, maybe change a few cases of `function()` to `function ()` per https://gist.github.com/rayshan/1ac3fd6329c03124e483#general-code-format, but both are common in the code base.

PR #1662, opened 2016-01-05
Little Optimisation Substitution & Slot Components.

thibaultzanini:


rayshan:
FYI this was done in montage-contour fork as well

PR #1674, opened 2016-01-13
removes the support for the “distinct” feature from Montag.defineProperty()

marchant:
… property descriptors which was primarily used for automatically assigning an array/object/collection to new objects. The recommended pattern now is to either assign the collection in a constructor if the property is systematically required, or creating a getter and a matching private property for holding the value and lazily do:

    aProperty: {
        get: function() {
            return this._aProperty || (this. _aProperty = []);
        }
    }


thibaultzanini:
LGTM

PR #1678, opened 2016-01-22
Fix issue that prevented scrolling over buttons on touch devices

### LINKED ISSUE #1677: Montage buttons prevent scrolling
If you have an element that scrolls its overflow that also contains Montage buttons, attempting to scroll over the buttons will be interpreted as a button click, which will therefore render scrolling impossible.

Tested on latest Chrome for Android and Safari on iPad.

cdebost:
Fixes #1677 

Note: In order for scrolling over buttons to work, they must respond to touch up events, i.e. press instead of pressStart. This seems appropriate. The behavior on non-touch clicks, however, is still to activate the button on mouse down rather than mouse up.

### DIFF
--- a/ui/button.reel/button.js
+++ b/ui/button.reel/button.js
@@ -213,10 +213,6 @@ var Button = exports.Button = Control.specialize(/** @lends module:"montage/ui/n
                 // it scrolls and not the webpage
                 document.addEventListener("touchmove", this, false);
             }
-
-            if (!this._preventFocus) {
-                this._element.focus();
-            }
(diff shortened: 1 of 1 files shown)


thibaultzanini:
@cdebost Thanks for this PR, it's actually an known bug. @benoit is going to provide a PR with all the native control in montage asap that should fix this bug. So, maybe in order to avoid conflict we should close this PR.

cdebost:
@thibaultzanini Sounds good to me.

PR #1683, opened 2016-01-30
Add Environment and fix an issue with the event filtering on Chrome.

thibaultzanini:
- Initial commit for the Environment Object. …
Allow montage to know its running environment. (os, browser, mobile device…)

- Fix an issue with the event filtering on Chrome.
Add some logic to handle the fact Touch Events and simulated Mouse Events can have a different target and not the same position.

### DIFF
--- a/core/environment.js
+++ b/core/environment.js
@@ -55,6 +55,21 @@ var Environment = exports.Environment = Montage.specialize({
         }
     },
 
+    _isBrowserSupportPointerEvents: {
+        value: null
+    },
+
+    isBrowserSupportPointerEvents: {
+        get: function () {
+            if (this._isBrowserSupportPointerEvents === null) {
+                this._isBrowserSupportPointerEvents = !!(window.PointerEvent ||
+                    (window.MSPointerEvent && window.navigator.msPointerEnabled));
+            }
+
+            return this._isBrowserSupportPointerEvents;
+        }
+    },
+
     _isIOSDevice: {
         value: null
     },
--- a/core/environment.js
+++ b/core/environment.js
@@ -333,3 +333,5 @@ var Environment = exports.Environment = Montage.specialize({
     }
 
 });
+
+exports.currentEnvironment = new Environment();
--- a/core/environment.js
+++ b/core/environment.js
@@ -55,18 +55,15 @@ var Environment = exports.Environment = Montage.specialize({
         }
     },
 
-    _isBrowserSupportPointerEvents: {
+    _supportsPointerEvents: {
         value: null
     },
 
-    isBrowserSupportPointerEvents: {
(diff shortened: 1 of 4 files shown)


rayshan:
Thanks for including `isCordova`!

thibaultzanini:
@rayshan @benoitmarchant PR updated

core/environment.js
```diff
@@ -0,0 +1,352 @@
...
+    },
+
+    _isWKWebView: {
+        value: null
+    },
+
+    isWKWebView: {
+        get: function () {
+            if (this._isWKWebView === null) {
+                this._isWKWebView = this.isIOSDevice && !!window.indexedDB;
```
marchant: This is penalizing every read, better to do:

return this._isWKWebView || (this._isWKWebView = this.isIOSDevice && !!window.indexedDB);

This applies to all methods here as much as possible

core/environment.js
```diff
@@ -0,0 +1,352 @@
...
+            return this._device;
+        }
+    },
+
+    _isBrowserSupportPointerEvents: {
+        value: null
+    },
+
+    isBrowserSupportPointerEvents: {
+        get: function () {
```
marchant: should be renamed "supportsPointerEvents"

core/event/event-manager.js
```diff
@@ -1642,7 +1638,7 @@ if (typeof window !== "undefined") { // client-side
                     return (this.memory[identifier] && (this.memory[identifier].size > 0));
                 },
                 storeEvent: function (mutableEvent) {
-                    var isBrowserSupportPointerEvents = this._isBrowserSupportPointerEvents,
+                    var isBrowserSupportPointerEvents = currentEnvironment.isBrowserSupportPointerEvents,
```
marchant: currentEnvironment should be a property on eventManager rather than a closure, storeEvent is used with high frequency so preferable to avoid a closure here

core/environment.js
thibaultzanini: `_isWKWebView` can be false. So, the best i can do here, it's to use a ternary operator but I'm not sure it's better than what we have already.

benoitmarchant: Right, when un-set is falsy, it's trickier, ternary is slightly faster

core/event/event-manager.js
thibaultzanini: This is actually a bug related to this commit:
https://github.com/montagejs/montage/commit/6004e8724594ae3c9d78ea5dc4613abb6dc92750
 
Here the context is different (pointerStorage) so the variable `isBrowserSupportPointerEvents` is always set to `undefined`.

thibaultzanini:
@rayshan @benoitmarchant PR updated

PR #1705, opened 2016-03-21
Remplace a binding by a couple getter/setter when creating a new Iteration.

thibaultzanini:


ui/repetition.reel/repetition.js
```diff
@@ -103,7 +103,34 @@ var Iteration = exports.Iteration = Montage.specialize( /** @lends Iteration.pro
...
+                this._active = active;
+
+                if (this.repetition) {
+                    if (active) {
+                        this.repetition.activeIterations.add(this);
+                    } else {
+                        this.repetition.activeIterations.delete(this);
+                    }
+
+                    this.repetition._addDirtyClassListIteration(this);
```
marchant: Should that be done regardless of active's value?

PR #1706, opened 2016-03-24
Fix the loader hiding itself too early

cdebost:


cdebost:
The loader causes issues in apps other than studio. Closing this until I get a better solution.

cdebost:
Replaced by https://github.com/montagejs/montage/pull/1708.

PR #1712, opened 2016-04-12
Raise a Press Event when it’s “safe” for touch devices.

thibaultzanini:
Knowing that, the simulated mouse events are not dispatched by the event manager, but they are still walking the dom. An issue could happen here when the positioning of elements is changing (z-index) after the press event has been raised, which could result to giving the focus to a wrong element. (see example in press-composer.info for user case)
- Updates specs.
- Add user case to the press-composer.info.

### DIFF
--- a/composer/press-composer.js
+++ b/composer/press-composer.js
@@ -736,7 +736,7 @@ var PressComposer = exports.PressComposer = Composer.specialize(/** @lends Press
 
                 if (this._needDispatchSafePress || event.defaultPrevented) {// no simulated event when a touchMove has been raised
                     var self = this,
-                        eventManager = this.element.ownerDocument.defaultView.defaultEventManager;
+                        eventManager = document.defaultView.defaultEventManager;
 
                     // Raise a Press event after the simulated mousedown event has been raised,
                     // in order to avoid elements that are not on the same layer  to get the focus.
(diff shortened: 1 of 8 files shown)


composer/press-composer.js
```diff
@@ -716,10 +733,38 @@ var PressComposer = exports.PressComposer = Composer.specialize(/** @lends Press
...
+                    // Raise a Press event after the simulated mousedown event has been raised,
+                    // in order to avoid elements that are not on the same layer  to get the focus.
+                    // @example: When a button is pressed in order to hide a overlay that it belong to.
+                    // Knowing that, the simulated mouse events are not dispatched by the event manager, but they are still
+                    // walking the dom. An issue could happen here when the positioning of elements is changing (z-index)
+                    // after the press event has been raised, which could result to giving the focus to a wrong element.
+                    // @example: @see press-composer.info
+                    window.nativeAddEventListener("mousedown", function _dispatchSafePress (mouseDownEvent) {
+                        if (touchEndTargetElement === mouseDownEvent.target ||
+                            eventManager._couldEmulatedEventHasWrongTarget(
```
thibaultzanini: _couldEmulatedEventHasWrongTarget -> _couldEmulatedEventHaveWrongTarget

rayshan:
This PR is tested and being actively used for a client project.

PR #1714, opened 2016-04-12
Fix an issue with the press composer and hiding the android keyboard.

thibaultzanini:
We need to keep the last element reached by a touchmove because we can't just rely on the position given by the touchEnd because in some cases, hiding the android keyboard for example, could result to change the press composer's element position between a touchStart and a touchEnd event. Therefore, getting the "wrong" target.


rayshan:
This PR is tested and being actively used for a client project.

marchant:
LGTM

PR #1731, opened 2016-06-09
Succession, Build-In/Out, ImageGallery 

marchant:


marchant:
tested for a while with multiple applications

PR #1736, opened 2016-07-06
updated latest button with promise

jonfu86:
Please review before merging!

### DIFF
--- a/ui/component.js
+++ b/ui/component.js
@@ -2209,6 +2209,9 @@ var Component = exports.Component = Target.specialize(/** @lends Component.proto
 
     _leTagArguments: {
         value: function () {
+            if (this === this.rootComponent) {
+                return;
+            }
             //jshint -W106
             var ownerModuleId = this.ownerComponent._montage_metadata.moduleId;
             var label = this._montage_metadata.label;


thibaultzanini:
The branch has conflicts can you fix it?

ui/button.reel/button.js
```diff
@@ -462,4 +495,4 @@ Button.addAttributes( /** @lends module:"montage/ui/native/button.reel".Button#
 */
     value: null
 
-});
+});
```
thibaultzanini: missing end line

marchant:
It's missing tests, examples in .info, what happens if multiple 2 promises are set one after the other and the first one resolves?

jonfu86:
updated with promise test and logic to handle multiple promises

ui/button.reel/button.js
```diff
@@ -4,7 +4,8 @@
     @module "montage/ui/native/button.reel"
 */
 var Control = require("ui/control").Control,
-    PressComposer = require("composer/press-composer").PressComposer;
+    PressComposer = require("composer/press-composer").PressComposer,
+    Dict = require("collections/dict");
```
marchant: Dict doesn't seems to be used, and generally shouldn't be going forward

ui/button.reel/button.js
```diff
@@ -83,20 +83,46 @@ var Button = exports.Button = Control.specialize(/** @lends module:"montage/ui/n
...
+        }
+    },
+
     /**
-        converter
         A Montage converter object used to convert or format the label displayed by the Button instance. When a new value is assigned to <code>label</code>, the converter object's <code>convert()</code> method is invoked, passing it the newly assigned label value.
         @type {Property}
         @default null
     */
+    converter: {
```
marchant: This already exists on Control, please remove the comments as it looks like it got left behind

ui/button.reel/button.js
```diff
@@ -133,9 +158,8 @@ var Button = exports.Button = Control.specialize(/** @lends module:"montage/ui/n
                 }
             }
 
-            this._label = String(value);
-
-            if (this.isInputElement) {
+            this._label = "" + value;
```
marchant: please keep what was there: this._label = String(value); it avoids creating a new string if value is already one

PR #1737, opened 2016-07-07
Fix an infinite loop under IE11 within the localizer.

thibaultzanini:
Fixes an infinite loop under IE11 when the key and the default message is undefined and remove an console.warm within the localizer.


marchant:
Tested in IE11, merging

PR #1738, opened 2016-07-14
Improvements with the injection of templates.

thibaultzanini:
- Named parameters are not mandatory anymore.
- Add support of named parameters and start parameter at the same level.


marchant:
Reviewed, merging

PR #1758, opened 2016-08-30
fixed tabIndex of thumbs and bug with value being set as _max

jonfu86:
Some bug fixes for slider.


jonfu86:
Tested manually and using automated test, no errors related to slider.

PR #1763, opened 2016-09-09
Migrate attribute value changes logic to a two levels Map.

thibaultzanini:
Preparatory works for buffered DOM changes.


ui/component.js
```diff
@@ -2904,14 +2904,53 @@ var Component = exports.Component = Target.specialize(/** @lends Component.proto
...
-             return this.__elementAttributeValues || (this.__elementAttributeValues = {});
-         }
-     },
+    __bufferedElementsAttributeValues: {
+        value: null
+    },
+
+    _bufferedElementsAttributeValues: {
+        get: function () {
+            return this.__bufferedElementsAttributeValues || (this.__bufferedElementsAttributeValues = new Map());
```
marchant: Should be WeakMap as we don't want to hold on "gone" elements

ui/component.js
```diff
@@ -2904,14 +2904,53 @@ var Component = exports.Component = Target.specialize(/** @lends Component.proto
...
+        }
+    },
+
+    bufferElementAttributeValue: {
```
marchant: let's rename this setElementAttributeValue, as by definition, Components buffer DOM Changes

ui/component.js
```diff
@@ -2904,14 +2904,53 @@ var Component = exports.Component = Target.specialize(/** @lends Component.proto
...
+            var bufferedElementsAttributeValues = this._bufferedElementsAttributeValues,
+                elementAttributeValues;
+
+            if ((elementAttributeValues = bufferedElementsAttributeValues.get(element))) {
+                elementAttributeValues.delete(attribute);
+            }
+        }
+    },
+
+    bufferedElementAttributeValue: {
```
marchant: getElementAttributeValue

ui/component.js
thibaultzanini: A weak map can't be iterated.

PR #1767, opened 2016-09-10
Add element attribute descriptors list

thibaultzanini:


ui/attribute-properties-descriptor.mjson
```diff
@@ -0,0 +1,160 @@
+{
```
marchant: Let's make it .json since that's what it is!

PR #1772, opened 2016-09-26
Fix wrong API documentation for removePathChangeListener method

pchaussalet:


core/core.js
```diff
@@ -1427,7 +1427,6 @@ var pathPropertyDescriptors = {
      * @function Montage#removePathChangeListener
      * @param {string} path
      * @param {object|function}
```
marchant: Let's add "handler" there the way path and beforeChange are

hthetiot:
lgtm

PR #1797, opened 2017-02-06
Predicate

marchant:


johnnykahalawai:
Quick question:

Why is it:

var predicate = new Predicate().initWithExpression("a.b");

and not:

var predicate = Predicate.initWithExpression("a.b");

PR #1805, opened 2017-03-21
implement pure JS montage/core/mini-url

hthetiot:
This is a pure implementation of URL.parse to get URL.resolve in order to provide the feature on env that do not support DOM.


hthetiot:
Instead may be MR can provide module 'url' like it could provide 'promise'

PR #1803, opened 2017-03-15
Upgrade test with montage-testing v2 stack 

hthetiot:
# Objectives

- migrate tests to jasmine 2 and karma
- migrate test from jasmine-node to pure NodeJS
- implement NodeJS test support expect for DOM related
- update to last mr#master
- update to last montage-testing#master
- update to last mop#master
- update to last mop-integration#master

# Requirement

- NO REGRESSION from #master in term of API or features.

Browser tests results:
screen shot 2017-04-12 at 9 26 21 am

NodeJs tests results:

screen shot 2017-04-12 at 9 22 31 am

PR #1806, opened 2017-03-23
implement proper Criteria serializeSelf

hthetiot:
I dont know if that required to match last montage-data.
May be not, will see how #1803 goes.

### DIFF
--- a/core/serialization/bindings.js
+++ b/core/serialization/bindings.js
@@ -22,12 +22,15 @@ Serializer.defineSerializationUnit("bindings", function (serializer, object) {
 
         var output = {};
 
-        if (("serializable" in input) && !input.serializable)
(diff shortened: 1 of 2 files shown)


core/serialization/bindings.js
```diff
@@ -22,12 +22,15 @@ Serializer.defineSerializationUnit("bindings", function (serializer, object) {
 
         var output = {};
 
-        if (("serializable" in input) && !input.serializable)
+        if (("serializable" in input) && !input.serializable) {
             continue;
+        } else if (!input.sourceSyntax) {
```
BnTejaede: Is this line required do other changes on the PR or is it just a general fix?  I don't have a problem with including it. I just want to make sure I understand if and how it fits.

PR #1811, opened 2017-04-05
Addresses a regression in iOS 10.3.1 around emulated events suppressi…

marchant:
…on, adds an optimization to only assess dispatching of emulated events for native events only


hthetiot:
Merged to run travis since we had none on master previously.

hthetiot:
See travis build here:
- https://travis-ci.org/montagejs/montage/builds/219022795

PR #1824, opened 2017-06-09
Add PipelineConverter

BnTejaede:


core/serialization/deserializer/self-deserializer.js
```diff
@@ -50,7 +50,7 @@ var SelfDeserializer = Montage.specialize( {
 
     getObjectByLabel: {
         value: function (label) {
-            this._context.getObject(label);
+            return this._context.getObject(label);
```
BnTejaede: @benoitmarchant 

This appeared to be just a typo. It was missing the return.

Issue #1820, opened 2017-05-05
Add property assignment to Bindings block that will replace properties


hthetiot:
Add property assignment to Bindings block that will replace properties
The ability to assign a value derived from another object in the serialization is missing. By adding a new way to make one-time assignment to bindings, we can later remove the properties block:

      "cnn": {
            "prototype": "ui/tv/tv-cnn.reel",
            "properties": {
                "element": {"#": "cnn"},
                "identifier": "cnn"
            },
            "bindings": {
                "isVisible": {"<-": "@owner.isVisible"}
            }
        },

Becomes

      "cnn": {
            "prototype": "ui/tv/tv-cnn.reel",
            "values": {
                "element": {"=": "#cnn"},
                "identifier": {"=": "cnn"},
                "foo": "FOO",
                "isVisible": {"<-": "@owner.isVisible"},
                "propA.probB": {"=": "@owner.blah.baz"}
            }
        },

cdebost:
@hthetiot Fixed by #1827?

thibaultzanini:
@cdebost Yes Fixed by #1827

PR #1827, opened 2017-06-14
Add property assignment to Bindings block that will replace properties

thibaultzanini:
- [x] Support for bindings within the values block
- [x] Support values block
- [x] Support one-time assignment to bindings
- [x] Deprecate bindings and properties blocks
- [x] Update serializer
- [x] Update serializer/deserializer/serialization Specs
- [x] Specs for one-time assignment to bindings
- [x] Cleaning pass
- [x] Fix object key that refer to another object 
- [x] Giving the path to a mjson file when specifying the object key should refer to a singleton 

#1820


hthetiot:
lgtm

hthetiot:
@thibaultzanini 
We old that for 17.1 or you think it's fully retro compatible ? If so set to 17.0.x for change logs and merge.

thibaultzanini:
@hthetiot It's backward compatible but I would like @benoitmarchant to check it as well and I think @cdebost wants to add a commit to this PR

cdebost:
@hthetiot In its current state this PR does not allow the user to define values onto an object deserialized from an .mjson, e.g.:

"mjsonProto": {
    "prototype": "foo.mjson",
    "values": {
        "bar": "baz"
    }
},

"mjsonObject": {
    "object": "foo.mjson",
    "values": {
        "bar": "baz"
    }
}

In both cases above "bar" will not be added to the two deserialized objects. My change would implement this.

My commit is ready to go but I'm unsure of whether to implement the feature above for the "object": "..." case. @thibaultzanini Let's discuss tomorrow when you have a moment.

benoitmarchant:
It's clearly needed for prototype, I think it should work for object as well. When a singleton exists, even programmatically, one someone mutate it, it impacts everyone, this would be no different except done through a serialization

cdebost:
@thibaultzanini Added my modifications for value deserializing onto .mjsons. Didn't include deprecated support for properties / bindings blocks because this is a new feature. Let me know what you think.

thibaultzanini:
@cdebost Looks good to me. @benoitmarchant your call :)

cdebost:
@thibaultzanini @marchant If there are no outstanding issues here could we get this PR merged? https://github.com/montagejs/montage/pull/1837 which removes circular dependencies and meta reference objects is going to depend on these changes.

marchant:
Let's go ahead and merge, the sooner we live with it, the sooner we may find potential issues, test passes, so let's merge

PR #1828, opened 2017-06-15
RangeController: Fix undefined self reference

cdebost:
RangeSelection uses self in its outer scope without having declared it. Not sure why RangeSelection.oldSwap is set in this way while RangeSelection.oldPush a few lines down does it differently.

Notes
* This bug was being masked in the browser (self points to window), so I enabled this spec for testing under a node env (and also Karma, though it also points self onto window).
* The change in range-controller-spec is for compatibility with Karma, but is semantically identical. Jasmine throws an error when running expect(arr).toContain(...) under Karma.
* Though self pointing to window under a browser or Karma environment prevents an exception, it is definitely not the correct logic (self.oldSwap is always undefined), and it's likely this bug was breaking swapping on range controllers in some cases.
* Not sure if self pointing to window is intended behavior. If not it's likely that some code in the montage core is accidentally defining self in the global scope.

### DIFF
--- a/core/range-controller.js
+++ b/core/range-controller.js
@@ -73,11 +73,7 @@ Object.defineProperty(_RangeSelection.prototype, "clone", {
         return this.slice();
     }
 });
-var oldSwap = self.swap;
-Object.defineProperty(_RangeSelection.prototype, "oldSwap", {
-    configurable: false,
-    value: observableArrayProperties.swap.value
-});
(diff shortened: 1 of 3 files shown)


hthetiot:
lgtm

hthetiot:
Right we should avoid self, may be change jshint if needed, the only globals in a module need to be same as node: require, module, export, global

PR #1833, opened 2017-07-12
[WIP] ES6 Features

cdebost:
⚠️ Do not merge.

List of features

- [ ] let & const
- [ ] arrow functions
- [ ] template strings
- [ ] classes


hthetiot:
@cdebost explain pls and set to 17.1 if you think this is short term delivery.

cdebost:
This is about refactoring the codebase to use the ES6 features listed
above. Definitely not short term nor particularly urgent, and won't be
merged onto master until we at the very least have a reliable webpack +
Babel setup because it breaks compatibility with some browsers we have to
support.

hthetiot:
ok thx @cdebost

hthetiot:
Should depend https://github.com/montagejs/montage/tree/commonjs branch

PR #1837, opened 2017-07-17
Remove circular references in meta/

### LINKED ISSUE #1832: Circular dependency in meta/*
Also have seen this issue as "Cannot read property 'specialized' of undefined". Both refer to PropertyDescriptor being undefined in either core/meta/derived-descriptor.js or core/meta/object-descriptor.js.

Can be simply reproduced with a blank montage app on #master that requires DerivedDescriptor and attempts to instantiate an instance.

Possibly caused by a circular dependency (DerivedDescriptor requires PropertyDescriptor which requires ObjectDescriptorReference which requires ObjectDescriptor which requires DerivedDescriptor).

Causes Studio to be unable to interact with a component's .mjson.

cdebost:
Fixes #1832. Depends on #1840.

The goal here is to use the new {"object": "xxx.mjson"} serialization syntax courtesy of @thibaultzanini to stop using meta reference objects (ObjectDescriptorReference, ModuleDescriptorReference etc.), which are the primary cause for circular references.

### DIFF
--- a/core/converter/bytes-converter.meta
+++ b/core/converter/bytes-converter.meta
@@ -11,16 +11,7 @@
         }
     },
     "converter_blueprint_reference": {
-        "prototype": "core/meta/blueprint-reference",
-        "values": {
-            "valueReference": {
-                "blueprintName": "Converter",
-                "prototypeName": "Converter",
-                "blueprintModule": {
-                    "%": "core/converter/converter.meta"
-                }
-            }
-        }
+        "object": "core/converter/converter.meta"
     },
     "root": {
         "prototype": "core/meta/module-blueprint",
--- a/ui/flow.reel/flow.meta
+++ b/ui/flow.reel/flow.meta
@@ -113,16 +113,7 @@
         }
     },
     "blueprint_component_reference": {
-        "prototype": "core/meta/blueprint-reference",
-        "values": {
-            "valueReference": {
-                "blueprintName": "Component",
-                "prototypeName": "Component",
-                "blueprintModule": {
-                    "%": "ui/component.meta"
-                }
-            }
-        }
+        "object": "ui/component.meta"
     },
     "root": {
         "prototype": "core/meta/module-blueprint",
--- a/ui/repetition.reel/repetition.meta
+++ b/ui/repetition.reel/repetition.meta
@@ -62,16 +62,7 @@
         }
     },
     "blueprint_component_reference": {
-        "prototype": "core/meta/blueprint-reference",
-        "values": {
-            "valueReference": {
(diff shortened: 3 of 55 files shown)


ui/flow.reel/flow.meta
```diff
@@ -1,7 +1,7 @@
 {
     "blueprint_flow_stride": {
         "prototype": "core/meta/property-blueprint",
```
marchant: Maybe worth updating property-blueprint to property-descriptor at the same time?

cdebost: Yep, will do that along with killing off meta reference objects. WIP.

cdebost:
@hthetiot What do you think about the milestone for this change? v17.1.x is supposed to be a backwards-compatible release, but this is a breaking change that completely removes core/meta/object-descriptor-reference and core/meta/blueprint-reference because they are the cause of circular requires. Though this PR updates montage itself to no longer use these references, apps that have .mjson models defined for montage data would break. ObjectDescriptorReference/BlueprintReference was the only way of creating references for a property descriptor's valueDescriptor property or a module descriptor's parent property prior to the "object": "..." syntax.

marchant:
@hthetiot, right, when we pull this in, looks like it would be more appropriate to go to 18.0.0

cdebost:
On second thought, I should really just leave the modules in, making sure none of the rest of the codebase is using them, and make a deprecation warning for them. That way we can merge in 17.x and remove them completely in 18.

ui/repetition.reel/repetition.meta
```diff
@@ -62,16 +62,7 @@
         }
     },
     "blueprint_component_reference": {
```
marchant: Also could be a good opportunity to rename .meta in .mjson

core/converter/bytes-converter.meta
```diff
@@ -11,16 +11,7 @@
...
-                "prototypeName": "Converter",
-                "blueprintModule": {
-                    "%": "core/converter/converter.meta"
-                }
-            }
-        }
+        "object": "core/converter/converter.meta"
     },
     "root": {
         "prototype": "core/meta/module-blueprint",
```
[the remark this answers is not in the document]
cdebost: There are a lot of places where the blueprint -> objectDescriptor renaming needs to be done. I'm going to do all of that in a separate PR, so this one does not get polluted with changes.

ui/repetition.reel/repetition.meta
cdebost: Same as above.

marchant:
Ok, reviewed, and yes, we'll release that under Montage 18

PR #1841, opened 2017-07-24
Remove logging & optimize PipelineConverter

BnTejaede:
- Remove logging pushed by mistake

- Refactor PipelineConverter to use a pointer instead of array.slice()


hthetiot:
lgtm

PR #1844, opened 2017-08-03
WebSocket improvements

cdebost:
* Removes _isMessagePending, which would block a websocket from sending a message until it has received a response.
* Removes a race condition where a websocket could be duplicated while connecting.


marchant:
LGTM

PR #1849, opened 2017-08-21
Fixes a resizing bug in Flow

thibaultzanini:
a precision logic error
⚠️  need montage-testing to be updated first
https://github.com/montagejs/montage-testing/pull/32


hthetiot:
merged montagejs/montage-testing#32 and restarted travis in here.

PR #1852, opened 2017-08-25
Deserializer Improvements

### LINKED ISSUE #1834: Remove didCreate() API
The Montage Reviver calls didCreate() or the constructor() on a newly instantiated Montage object, depending on which is available. We should completely remove the didCreate() API, as it is called the exact same way as the constructor, adding no extra value. Additionally, defining a didCreate function on a Montage prototype will also unintuitively cause the constructor function to not be called when instantiated.

cdebost:
Fixes #1834.

- [x] Decouple deserializer from reviver
- [x] Add deserializer caching
- [x] Use new instead of Object.create() for constructor modules, remove didCreate()
- [x] Make `deserializeSelf` implementations reentrant so they can be called multiple times on the same instance.
- [x] Measure performance:

#### Performance Analysis

screen shot 2017-09-20 at 5 42 56 pm

When deserializing only JS modules, the performance difference with this change is negligible. There is a very small exponential speedup (about 5% faster over a sample size of 1.6 million deserializations) due to the change from Object.create() to new.

screen shot 2017-09-07 at 10 58 08 am

We see massive performance gains when deserializing nested .mjson files thanks to deserialization caching. The deserializer is able to run this test with a better-than-linear time complexity, whereas the master implementation is exponential.

### DIFF
--- a/core/meta/model.js
+++ b/core/meta/model.js
@@ -64,17 +64,23 @@ var Model = exports.Model = Montage.specialize( /** @lends Model.prototype # */ 
     deserializeSelf: {
         value: function (deserializer) {
             var value = deserializer.getProperty("version");
-            if (value !== undefined) {
+            if (value !== void 0) {
                 this.version = value;
             }
 
-            this._name = deserializer.getProperty("name");
+            value = deserializer.getProperty("name");
+            if (value !== void 0) {
+                this._name = value;
+            }
             //copy contents into the objectDescriptors array
             value = deserializer.getProperty("objectDescriptors") || deserializer.getProperty("blueprints");
             if (value) {
                 this._objectDescriptors = value;
             }
-            this.modelInstanceModuleId = deserializer.getProperty("objectModelModuleId") || deserializer.getProperty("binderModuleId");
+            value = deserializer.getProperty("objectModelModuleId") || deserializer.getProperty("binderModuleId");
+            if (value !== void 0) {
+                this.modelInstanceModuleId = value;
+            }
         }
     },
 
--- a/core/serialization/deserializer/montage-deserializer.js
+++ b/core/serialization/deserializer/montage-deserializer.js
@@ -1,14 +1,10 @@
 var Montage = require("../../core").Montage,
-    MontageInterpreter = require("./montage-interpreter").MontageInterpreter,
+    MontageContext = require("./montage-interpreter").MontageContext,
     MontageReviver = require("./montage-reviver").MontageReviver,
     Map = require("collections/map").Map,
     deprecate = require("../../deprecate");
 
 var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
-
-    _interpreter: {
-        value: null
-    },
 
     _serializationString: {
         value: null
@@ -29,9 +25,9 @@ var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
     init: {
         value: function (serializationString, _require, objectRequires, locationId, moduleContexts) {
             this._serializationString = serializationString;
+            this._require = _require;
             moduleContexts = moduleContexts || new Map();
-            this._interpreter = new MontageInterpreter().init(_require,
-                new MontageReviver().init(_require, objectRequires, locationId, moduleContexts));
+            this._reviver = new MontageReviver().init(_require, objectRequires, locationId, moduleContexts);
 
             return this;
         }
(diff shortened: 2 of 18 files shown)


core/meta/model.js
```diff
@@ -64,17 +64,23 @@ var Model = exports.Model = Montage.specialize( /** @lends Model.prototype # */
...
+            if (value !== void 0) {
+                this._name = value;
+            }
             //copy contents into the objectDescriptors array
             value = deserializer.getProperty("objectDescriptors") || deserializer.getProperty("blueprints");
             if (value) {
                 this._objectDescriptors = value;
             }
-            this.modelInstanceModuleId = deserializer.getProperty("objectModelModuleId") || deserializer.getProperty("binderModuleId");
+            value = deserializer.getProperty("objectModelModuleId") || deserializer.getProperty("binderModuleId");
```
marchant: It would be a good idea to add comments to explain that now a deserialization can be re-entring which is why the deserialization logic has to be adapted. This is something we should document and explain as well

core/serialization/deserializer/montage-deserializer.js
```diff
@@ -27,39 +23,62 @@ var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
...
+        value: function (serialization, _require, objectRequires, locationId) {
+            if (typeof serialization === "string") {
+                this._serializationString = serialization;
+            } else {
+                this._serializationString = JSON.stringify(serialization);
+            }
+            this._require = _require;
+            this._locationId = locationId;
+            this._reviver = new MontageReviver().init(_require, objectRequires,
+                this._childConstructor.bind(this));
```
marchant: This will create a new function everytime a deserializer is created, can't we use call() or pass this as an extra argument instead?

core/serialization/deserializer/montage-deserializer.js
```diff
@@ -76,9 +95,39 @@ var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
 
     preloadModules: {
         value: function () {
-            var serialization = JSON.parse(this._serializationString);
+            var serialization = JSON.parse(this._serializationString),
```
marchant: JSON.parse(this._serializationString) is also executed in deserialize(), can't that be done only once?

core/serialization/deserializer/montage-deserializer.js
```diff
@@ -76,9 +95,39 @@ var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
...
+                reviver = this._reviver,
+                moduleLoader = reviver.moduleLoader,
+                object,
+                locationId,
+                locationDesc,
+                module,
+                promises = [];
+
+            for (var label in serialization) {
+                if (serialization.hasOwnProperty(label)) {
```
marchant: Since that loop only does something when label is a property on serialization directly, it would probably be faster to just loop on Object.keys() which returns an array of a given object's own enumerable properties.

core/serialization/deserializer/montage-reviver.js
```diff
@@ -1023,8 +990,9 @@ var MontageReviver = exports.MontageReviver = Montage.specialize(/** @lends Mont
                     typeof reviver[methodName] === "function" &&
                         methodName.substr(0, 5) === "revive"
                 ) {
-                    if (typeof customObjectRevivers[methodName] === "undefined") {
-                        customObjectRevivers[methodName] = reviver[methodName].bind(reviver);
+                    customObjectReviver = this.customObjectRevivers.get(methodName);
+                    if (typeof customObjectReviver === "undefined") {
+                        customObjectReviver = reviver[methodName].bind(reviver);
```
marchant: .bind() create a new function, could we find a way to avoid doing so? That could boost perfs

core/serialization/deserializer/montage-deserializer.js
cdebost: Something strange is happening. If we just do the parse once in the init instead of doing it at the beginning of preloadModules() and deserialize(), a bunch of tests fail with "Converting circular structure to JSON" errors. I think it's because in the case where the deserializer is passed an object instead of a string, the result of JSON.parse-ing the JSON.stringify-ed object is different from the original object. So for now when the deserializer is initialized with an object, it has to stringify() it so it can then parse() it and avoid these errors. This shouldn't be the case, but I couldn't figure out the solution yet, we should shelve this change for another PR.

core/serialization/deserializer/montage-reviver.js
cdebost: Hmm...
`customObjectReviver` is a local variable and isn't used again after it is assigned to `reviver[methodName].bind(reviver)`, so isn't that assignment actually pointless?

cdebost: Actually, I think this bind has to stay, because we're grabbing a method from a user-supplied object.
Anyway, custom object revivers aren't actually used anywhere in montage or filament/palette, so we won't see a performance boost from removing bind.

core/meta/model.js
cdebost: Agreed, but deserializeSelf is not a concretely defined api on a specific type. The deserializer just looks for a `deserializeSelf` method on whatever object it loads, and then expects it to work the way it wants. `deserializeSelf` isn't specific to the Montage type so I'm not sure where we would comment this in the code.
For general documentation, we should probably have some doc on the website that explains the serialization process and how to hook into it using serializeSelf and deserializeSelf. I think that would be good to have but outside the scope of this PR.

marchant: Agreed, we'll need to write a technical doc

marchant:
LGTM! Merging

Issue #1851, opened 2017-08-25
Some converters need to be singleton.


thibaultzanini:
For performance purpose some stateless converters should be a singleton, such as the `trimConverter`.
Moreover, we could use the package `lodash` for some of our converters.

thibaultzanini:
Wait for https://github.com/montagejs/montage/issues/1851

PR #1853, opened 2017-08-29
Converters singleton mecanism + new converters

thibaultzanini:
Fix: #1851 

Add converters: 
- CamelCaseConverter
- KebabCaseConverter
- SnakeCaseConverter.


core/converter/camel-case-converter.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+var CamelCaseConverter = exports.CamelCaseConverter = Converter.specialize({
+    convert: {
+        value: function (v) {
+            return camelCase(v);
+        }
+    }
+});
+
+Object.defineProperty(exports, 'defaultCamelCaseConverter', {
```
marchant: I'm not sure how that defaultCamelCaseConverter pattern works. From what I understand, if someone does:
var CamelCaseConverter = require("montage/core/converter/camel-case-converter").CamelCaseConverter,
camelCaseConverterA = new CamelCaseConverter(),
camelCaseConverterB = new CamelCaseConverter();

then camelCaseConverterA !== camelCaseConverterB. If it's going to be a singleton I think it has to be by default, meaning the constructor needs to return the singleton, always. If one has to use 
var CamelCaseConverter = require("montage/core/converter/camel-case-converter"). defaultCamelCaseConverter

for that, then it wouldn't work as seamlessly in serialization and would likely be ignored

thibaultzanini: The problem is the following, at the moment we declare a converter within the serialization as such:
```
"converter": {
    "prototype": "montage/core/converter"
}
``` 
Which create a new instance of a converter, but we all agree recently (the work on .mjson files) that the key `prototype` should be used for new instance and `object` for object, singleton...
Plus, it could be strange for a developer to always get a singleton after calling the `new` operator.

I believe something should probably be done at the deserialization time.

core/converter/camel-case-converter.meta
```diff
@@ -0,0 +1,34 @@
...
+                "blueprintName": "Converter",
+                "prototypeName": "Converter",
+                "blueprintModule": {
+                    "%": "core/converter/converter.meta"
+                }
+            }
+        }
+    },
+    "root": {
+        "prototype": "core/meta/module-blueprint",
```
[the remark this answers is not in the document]
thibaultzanini: oops should be an `module-object-descriptor` right? 
Maybe I could move the logic from`.meta` to `.mjson` files and remove all `.meta` files with this PR?

core/converter/camel-case-converter.js
marchant: In software engineering, the singleton pattern is a software design pattern that:
- Ensure that only one instance of a class is created.
- Provide a global point of access to the object.

Since these objects are stateless, which is the root idea behind making them singletons, there is no reason to burden the developer with that knowledge when we have a way, creating a singleton object and returning it in the constructor, only when asked. To the outside the developer is getting an instance of the class he wants and that's all he needs to know

core/converter/camel-case-converter.meta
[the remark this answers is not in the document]
marchant: Yes, that would be a good opportunity to do so

core/converter/bytes-converter.mjson
```diff
@@ -1,34 +1,34 @@
...
                 "@": "root"
             },
             "valueType": "number",
             "helpKey": ""
         }
     },
-    "converter_blueprint_reference": {
-        "prototype": "core/meta/blueprint-reference",
+    "converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" } instead of an instance of an object-descriptor-reference.

core/converter/camel-case-converter.mjson
```diff
@@ -0,0 +1,34 @@
+{
+    "converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" }.

core/converter/currency-converter.mjson
```diff
@@ -1,45 +1,45 @@
...
                 "@": "root"
             },
             "valueType": "string",
             "helpKey": ""
         }
     },
-    "number_converter_blueprint_reference": {
-        "prototype": "core/meta/blueprint-reference",
+    "number_converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/number-converter.mjson" }.

core/converter/date-converter.mjson
```diff
@@ -1,45 +1,45 @@
...
                 "@": "root"
             },
             "valueType": "string",
             "helpKey": ""
         }
     },
-    "converter_blueprint_reference": {
-        "prototype": "core/meta/blueprint-reference",
+    "converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" }.

core/converter/expression-converter.mjson
```diff
@@ -1,12 +1,12 @@
 {
-    "converter_object-descriptor_reference": {
+    "converter_descriptor_reference": {
         "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" }.

core/converter/invert-converter.mjson
```diff
@@ -1,29 +1,29 @@
 {
-    "converter_blueprint_reference": {
-        "prototype": "core/meta/blueprint-reference",
+    "converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" }.

core/converter/kebab-case-converter.mjson
```diff
@@ -0,0 +1,34 @@
+{
+    "converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" }.

core/converter/lower-case-converter.mjson
```diff
@@ -1,29 +1,29 @@
 {
-    "converter_blueprint_reference": {
-        "prototype": "core/meta/blueprint-reference",
+    "converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" }.

core/converter/new-line-to-br-converter.mjson
```diff
@@ -1,29 +1,29 @@
 {
-    "converter_blueprint_reference": {
-        "prototype": "core/meta/blueprint-reference",
+    "converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" }.

core/converter/number-converter.mjson
```diff
@@ -1,89 +1,89 @@
...
                 "@": "root"
             },
             "valueType": "boolean",
             "helpKey": ""
         }
     },
-    "converter_blueprint_reference": {
-        "prototype": "core/meta/blueprint-reference",
+    "converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" }.

core/converter/snake-case-converter.mjson
```diff
@@ -0,0 +1,34 @@
+{
+    "converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" }.

core/converter/trim-converter.mjson
```diff
@@ -1,29 +1,29 @@
 {
-    "converter_blueprint_reference": {
-        "prototype": "core/meta/blueprint-reference",
+    "converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" }.

core/converter/upper-case-converter.mjson
```diff
@@ -1,29 +1,29 @@
 {
-    "converter_blueprint_reference": {
-        "prototype": "core/meta/blueprint-reference",
+    "converter_descriptor_reference": {
+        "prototype": "core/meta/object-descriptor-reference",
```
cdebost: Deprecated. Use { "object": "core/converter/converter.mjson" }.

core/converter/camel-case-converter.js
thibaultzanini: @corentin 
I don't like that the constructors for the converters return the singleton. If I call `new` on a Montage object I want a new instance, not a singleton. A singleton should be exposed in a constructor property, or for serialization exposed as a separate object in the exports that we could reference with something like
```json
{
  "object": "core/converter/camel-case-converter[instance]"
}
```

thibaultzanini: @marchant I don't like it either I think It's a little bit confusing

Corentin:
Wrong person

marchant:
Reviewed

PR #1860, opened 2017-09-14
Allow to set directly values without paths when using the one time assignment syntax.

thibaultzanini:
Specs updated.


hthetiot:
lgtm

PR #1870, opened 2017-09-21
Add Object.assign shim

cdebost:


hthetiot:
We need this is this good to have it and be able to start using it.

PR #1876, opened 2017-10-02
Lint and merge master into features/data branch

cdebost:
This is also a merge. Changes not already in master start at the [lint] commits.

Remaining errors:

- [x] `data/service/persistent-data-service.js: line 487, col 11, Duplicate key 'storage'.` PersistentDataService has two `storage` properties. One is a getter for the private _storage property, the other is a function that returns a rejected promise.

### DIFF
--- a/data/service/raw-data-service.js
+++ b/data/service/raw-data-service.js
@@ -643,7 +643,6 @@ exports.RawDataService = DataService.specialize(/** @lends RawDataService.protot
                     stream.addData([mappedObject]);
                 });
             }
-
         }
     },
 
--- a/data/service/raw-data-service.js
+++ b/data/service/raw-data-service.js
@@ -639,9 +639,15 @@ exports.RawDataService = DataService.specialize(/** @lends RawDataService.protot
             // will be done in place to avoid creating any unnecessary array.
             for (i = 0, n = records && records.length; i < n; i += 1) {
                 iRecord = records[i];
+                /*jshint -W083*/
+                // Turning off jshint's function within loop warning because the
+                // only "outer scoped variable" we're accessing here is stream,
+                // which is a constant reference and won't cause unexpected
+                // behavior due to iteration.
                 this.addOneRawData(stream, iRecord, context, streamSelectorType).then(function (mappedObject) {
                     stream.addData([mappedObject]);
                 });
+                /*jshint +W083*/
             }
         }
     },
--- a/data/service/snapshot-service.js
+++ b/data/service/snapshot-service.js
@@ -111,30 +111,25 @@ exports.SnapshotService = Montage.specialize(/** @lends SnapshotService# */ {
         value: function(a, b) {
             var result = a === b;
             if (!result) {
-                if (typeof a === "object" && typeof b === "object") {
-                    
-                    // TODO WTF
-                    result = (!!a === !!b);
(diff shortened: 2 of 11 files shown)


hthetiot:
lgtm for basic lint, I have not reviewed logic.

data/service/snapshot-service.js
```diff
@@ -111,30 +111,25 @@ exports.SnapshotService = Montage.specialize(/** @lends SnapshotService# */ {
         value: function(a, b) {
             var result = a === b;
             if (!result) {
-                if (typeof a === "object" && typeof b === "object") {
-                    
-                    // TODO WTF
-                    result = (!!a === !!b);
```
cdebost: ;)

This checks if a and b have the same truthy values, but doing it this way isn't necessary. We already know from previous conditions that `a !== b` and `a !== undefined` and `b !== undefined`. The next part of the function uses Object.keys(), which needs a truthy (non-null) parameter, so we just need to check that a and b are both truthy. If they are both falsy objects then the `a === b` condition on the first line would have been true already.

data/service/raw-data-service.js
```diff
@@ -639,11 +639,16 @@ exports.RawDataService = DataService.specialize(/** @lends RawDataService.protot
             // will be done in place to avoid creating any unnecessary array.
             for (i = 0, n = records && records.length; i < n; i += 1) {
                 iRecord = records[i];
+                /*jshint -W083*/
+                // Turning off jshint's function within loop warning because the
+                // only "outer scoped variable" we're accessing here is stream,
+                // which is a constant reference and won't cause unexpected
+                // behavior due to iteration.
```
cdebost: The warning is not helpful in this case because we don't use any non-constant variables inside the function body, and I don't want to have to create extra functions just to make jshint happy.

hthetiot:
Here is the last issue:

```
data/service/persistent-data-service.js: line 487, col 11, Duplicate key 'storage'.
```

I did not found the duplicate on my initial lint.

cdebost:
@BnTejaede Do you know which implementation of `storage` in `persistent-data-service.js` is the correct one? Here are the two implementations:

```

     * returns a Promise that resolves to the object used by the service to
     * store data. This is meant to be an abstraction of the "Database"
     *
     * @returns {Promise}
     */

    _storage: {
        value: undefined
    },
    storage: {
        get: function() {
            return this._storage || (this._storage = Promise.reject(new Error('Needs to be implemented by sub classes')));
        }
```

```

     * Returns a Promise for the storage used to store objects
     * described by all objectDescriptors in the service's model.
     *
     * @argument {Model} stream
     * @returns {Promise}
     */
   storage: {
        value: function(model) {
            return Promise.reject();
        }
```

BnTejaede:
@cdebost I do not know for certain. That property does not exist in offline-data-service, which is the predecessor of persistent-data-service. @marchant is more likely to know.

rdwrite:
@cdebost: The first "storage" is the one to save (along with the private _storage that precedes it). The 2nd one is going away in the version of PersistentDataService I'm working on.

cdebost:
Cool. Thanks @rdwrite. I'm going to merge this now onto features/data in the interest of having a passing build.

rdwrite:
Thanks @cdebost!

PR #1877, opened 2017-10-03
Include every dependencies mentioned in mjson/meta files when mopping an app.

thibaultzanini:


hthetiot:
lgtm

Will check that in commonjs branch also

BnTejaede:
LGTM 

I verified that this causes mop to include the entire .mjson tree in the bundles.

PR #1878, opened 2017-10-06
Fix PropertyDescriptor._overridePropertyWithDefaults()

BnTejaede:
@cdebost 

The loop in overridePropertyWithDefaults method was not exiting properly. It checks for `!value`, but the loop was assigning `this[objectKey]` directly rather than assign `value`. So any property with more than one property name ended up with the value of the last property name. 

Also, `this._overridePropertyWithDefaults()` requires at least 3 arguments, but the call for only passed 2. The simplest fix was to call `this._overridePropertyWithDefaults(deserializer, "cardinality", "cardinality");`, but I also added a commented out snippet in `_overridePropertyWithDefaults` that could solve the problem in a more robust way. 

```javascript
// [TJ] Prospective code
// if (propertyNames.length === 0) {
 //     propertyNames = [objectKey];
 // }
```

Rather than require the developer to pass in the same string twice, we could assume that the key in the serialization is the same as the property key that we assign it to. Small thing, but it could make the deserializeSelf read a bit easier.

### DIFF
--- a/core/meta/property-descriptor.js
+++ b/core/meta/property-descriptor.js
@@ -125,10 +125,12 @@ exports.PropertyDescriptor = Montage.specialize( /** @lends PropertyDescriptor# 
             if (value !== void 0) {
                 this._owner = value;
             }
-            this._overridePropertyWithDefaults(deserializer, "cardinality");
+            this._overridePropertyWithDefaults(deserializer, "cardinality", "cardinality");
+          
             if (this.cardinality === -1) {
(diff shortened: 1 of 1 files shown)


cdebost:
lgtm. I like the solution for only passing in one key argument, is there a reason why it should be left commented out?

BnTejaede:
Nope. The only reason is whether you approved of it. :) Please feel free to
uncomment! I can also do it later this afternoon.

PR #1880, opened 2017-10-09
Tentative tweak to make improvements to the use of requestAnimation…

marchant:
…Frame

- removes use of workaround for lack of native requestAnimationFrame which exists in IE10

### DIFF
--- a/composer/press-composer.js
+++ b/composer/press-composer.js
@@ -867,7 +867,7 @@ var PressComposer = exports.PressComposer = Composer.specialize(/** @lends Press
                 var self = this;
 
                 this._longPressTimeout = setTimeout(function () {
-                    self._dispatchLongPress();
+                    self._dispatchLongPress(event.target);
                 }, this._longPressThreshold);
             }
         }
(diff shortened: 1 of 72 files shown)


hthetiot:
Why removing the feature detection in place ? Fallback should still be available.

hthetiot:
The only issue I have is with removing the shim and not supporting prefixed webkit.
- http://caniuse.com/#feat=requestanimationframe

Cc @BnTejaede can you confirm PDC app is ok without this shim on iOS notably.

BnTejaede:
@hthetiot @marchant I cannot load Contour in mobile Safari with the PR, but I can with Montage#master. I do not have the bandwidth to investigate further at the moment, but it appears that Harold may be right about the webkit prefixes.

hthetiot:
@marchant need someking of rebase on master, I dont know why this push is weird.

hthetiot:
need re-push cause PR is a mess.
Also see fix for regression #1923
Sorry.

PR #1881, opened 2017-10-09
Add targetElement property to longPress event

BnTejaede:
Unlike "press",  "longPress" does not provide which element was pressed to generate the event. It only provides the component to which the listener was attached. This prevents the parent component from using event delegation if it would like to use long press.

For example, say you have a directory viewer that works similar to macOS finder and supports two interactions: 
1. Select a list item on press.
2. Enter "edit mode" for a list item on long press. 

The following handlePress works with the current API. handleLongPress does not: 
```javascript
handlePress: {
    value: function (event) {
        event.stopPropagation();
        var file = event.targetElement.component.file;
        this.selectFile(file);
    }
},

handleLongPress: {
    value: function (event) {
        event.stopPropagation();
        var file = event.targetElement.component.file;
        this.editFile(file);
    }
},
```

As for the changes, my understanding of the PressComposer is that longPress will be dispatched only if the cursor does not move for the duration of the press. Therefore, the targetElement of a longPress is the same as the target of the original pointerDown.

### DIFF
--- a/composer/press-composer.js
+++ b/composer/press-composer.js
@@ -867,7 +867,7 @@ var PressComposer = exports.PressComposer = Composer.specialize(/** @lends Press
                 var self = this;
 
                 this._longPressTimeout = setTimeout(function () {
-                    self._dispatchLongPress();
+                    self._dispatchLongPress(event.target);
                 }, this._longPressThreshold);
             }
         }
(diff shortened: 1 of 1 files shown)


hthetiot:
lgtm

Issue #1887, opened 2017-10-29
Audit ObjectDescriptor.propertyDescriptors 


tejaede:
ObjectDescriptor has a couple of anti-patterns in it's management of propertyDescriptors that require some attention.  Most notably, ObjectDescriptor.propertiesDescriptors.get() creates a new array each time it is called.  https://github.com/montagejs/montage/blob/master/core/meta/object-descriptor.js#L381-L390

Dependencies: 
- [x] Make propertyDescriptorsTable a Map https://github.com/montagejs/montage/issues/1890
- [x] Add caching to ObjectDescriptor.propertiesDescriptors https://github.com/montagejs/montage/issues/1891

PR #1889, opened 2017-10-29
Add support for ObjectDescriptor & DataMapping inheritance

tejaede:
Add the ability to determine which subclass of a parent class to use for a particular rawData object.

Additionally, this PR includes: 
1. Fix in ObjectDescriptor.propertyDescriptors to allow a child ObjectDescriptor to override a propertyDescriptor of it's parent.
2. Removal of a couple hundred lines of commented out code from RawDataService

TODO Before Merge:
- [X] Optimize loop patterns to avoid array.length references when unnecessary.
- [x] Audit ObjectDescriptor.propertyDescriptors to clarify implementation. https://github.com/montagejs/montage/issues/1887
- [x] Traverse prototype chain while mapping


data/service/raw-data-service.js
```diff
@@ -1118,6 +893,69 @@ exports.RawDataService = DataService.specialize(/** @lends RawDataService.protot
...
+     * @param {Array<RawDataTypeMapping>} mappings 
+     */
+    _registerRawDataTypeMappings: {
+        value: function (mappings) {
+            var mapping, type, 
+                i, n;
+
+            for (i = 0, n = mappings ? mappings.length : 0; i < n; i++) {
+                mapping = mappings[i];
+                type = mapping.type;
```
marchant: type itself isn't used, but type.parent is, 3 times, so that's probably what should be put in the local variable

core/meta/object-descriptor.js
```diff
@@ -380,15 +384,30 @@ var ObjectDescriptor = exports.ObjectDescriptor = Montage.specialize( /** @lends
      */
     propertyDescriptors: {
         get: function () {
-            var propertyDescriptors = [];
-            propertyDescriptors = propertyDescriptors.concat(this._propertyDescriptors);
+            var propertyDescriptors = this._propertyDescriptors.slice();
```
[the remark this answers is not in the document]
BnTejaede: Agreed. I created a general case related to _propertyDescriptors. https://github.com/montagejs/montage/issues/1887

I see you created a specific one for _propertyDescriptorsTable. https://github.com/montagejs/montage/issues/1890 


So I propose one of two paths forward: 
1. Re-purpose case https://github.com/montagejs/montage/issues/1887 to specifically address invalidating the cache as you suggest
2. Create a new case to address invalidating the cache and treat https://github.com/montagejs/montage/issues/1887 as an epic that has the new case and https://github.com/montagejs/montage/issues/1890  as sub-cases. 

What do you think?

core/meta/object-descriptor.js
```diff
@@ -363,13 +363,15 @@ var ObjectDescriptor = exports.ObjectDescriptor = Montage.specialize( /** @lends
 
     _handlePropertyDescriptorsRangeChange: {
         value: function (plus, minus, index) {
-            var i, n;
-            for (i = 0, n = minus.length; i < n; ++i) {
-                minus[i]._owner = null;
+            var descriptor, i, n;
```
rdwrite: Unused `n`?

core/meta/object-descriptor.js
```diff
@@ -363,13 +363,15 @@ var ObjectDescriptor = exports.ObjectDescriptor = Montage.specialize( /** @lends
...
+            for (i = 0; (descriptor = minus[i]); ++i) {
+                descriptor._owner = null;
+                delete this._propertyDescriptorsTable[descriptor.name];
             }
 
-            for (i = 0, n = plus.length; i < n; ++i) {
-                plus[i]._owner = plus[i]._owner || this;
+            for (i = 0, n; (descriptor = plus[i]); ++i) {
```
rdwrite: Unused `n`?

core/meta/object-descriptor.js
BnTejaede: Good catch! I have other not-yet-pushed changes related to that method and I will be be sure to remove that var.

core/meta/object-descriptor.js
BnTejaede: Same as above.

core/converter/pipeline-converter.js
```diff
@@ -45,6 +73,21 @@ exports.PipelineConverter = Converter.specialize({
         value: undefined
     },
 
+   
+
+    _arrayToSet: {
```
marchant: Collections offers Set.from(anArray) that does that, let's use it?

core/meta/object-descriptor.js
```diff
@@ -357,35 +358,121 @@ var ObjectDescriptor = exports.ObjectDescriptor = Montage.specialize( /** @lends
     },
 
 
-    _propertyDescriptors: {
-        value: null
-    },
+    _arrayToSet: {
```
marchant: Similar code is in pipeline-converter.js, we can use Set.from() that Collections provide

core/meta/object-descriptor.js
```diff
@@ -357,35 +358,121 @@ var ObjectDescriptor = exports.ObjectDescriptor = Montage.specialize( /** @lends
...
+     * @property {Array<PropertyDescriptor>} value
      */
     propertyDescriptors: {
         get: function () {
-            var propertyDescriptors = [];
-            propertyDescriptors = propertyDescriptors.concat(this._propertyDescriptors);
-            if (this.parent) {
-                propertyDescriptors = propertyDescriptors.concat(this.parent.propertyDescriptors);
+            this._preparePropertyDescriptorsCache();
+            return this._propertyDescriptors || [];
```
marchant: sounds funny to return an empty array every time and not assign it to this._propertyDescriptors the first time if this._propertyDescriptors isn't  a thing yet ?

core/meta/object-descriptor.js
```diff
@@ -357,35 +358,121 @@ var ObjectDescriptor = exports.ObjectDescriptor = Montage.specialize( /** @lends
...
+        }
+    },
+
+    _validParentPropertyDescriptors: {
+        value: function () {
+            var parentPropertyDescriptors = this.parent.propertyDescriptors.slice(),
```
marchant: I don't see why we're making a copy (slice) of this.parent.propertyDescriptors here, we're looking on content and adding objects to something else

core/meta/property-descriptor.js
```diff
@@ -370,22 +370,22 @@ exports.PropertyDescriptor = Montage.specialize( /** @lends PropertyDescriptor#
                 deprecate.deprecationWarningOnce("valueDescriptor reference via ObjectDescriptorReference", "direct reference via object syntax");
                 return this._valueDescriptorReference.promise(this.require);
             } else {
-                return Promise.resolve(this._valueDescriptorReference);
+                return this._valueDescriptorReference && Promise.resolve(this._valueDescriptorReference);
```
marchant: What happens when this._valueDescriptorReference is false? undefined is returned when a promise is expected?

core/converter/pipeline-converter.js
tejaede: Good catch! At one point I noticed the constructor wasn't working in IE and it didn't occur to me to require collections/set so this was solution I went with. I'll update this.

tejaede: Realized I misread your comment. Collections adds support for `new Set(anArray)`, so I'd prefer to go with that over `Set.from(anArray)` because the constructor is a part of the spec. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

core/meta/object-descriptor.js
tejaede: Will update to use `new Set(anArray)` as well.

core/meta/object-descriptor.js
tejaede: Agreed. It serves a purpose though so we'll need a new strategy for that. 

The issue is that we can't build out the propertyDescriptor caches until every propertyDescriptor has a name. You can see this at work in `_preparePropertyDescriptorsCache`

https://github.com/montagejs/montage/blob/master/core/meta/object-descriptor.js#L408-L435

At the time I avoided using a special flag to gate whether we should try to build the cache instead of using `this._propertyDescriptors` like I do here: 


```javascript
            if (!this._propertyDescriptors)  {
                
                for (i = 0, n = ownDescriptors.length; i < n && isReady; ++i) {
                    descriptor = ownDescriptors[i];
                    isReady = !!(descriptor && descriptor.name);
                }
```

Though on further reflection it does seem wasteful to repeatedly create a new array only in order to avoid one boolean. So we could instead have a flag such as `_propertyDescriptorsAreCached` or `_propertyDescriptorsAreReady`. Then that would change `_preparePropertyDescriptorsCache` to the following: 

```javascript
_preparePropertyDescriptorsCache: {
    value: function () {
        var ownDescriptors = this._ownPropertyDescriptors,
            isReady = true,
            descriptor, i, n;
        
        if (!this._propertyDescriptorsAreCached)  {
            
            for (i = 0, n = ownDescriptors.length; i < n && isReady; ++i) {
                descriptor = ownDescriptors[i];
                isReady = !!(descriptor && descriptor.name);
            }

            if (isReady) {
                this._propertyDescriptorsAreCached = true;
                this._propertyDescriptorsTable.clear();
                for (i = 0, n = ownDescriptors.length; i < n && isReady; ++i) {
                    descriptor = ownDescriptors[i];
                    descriptor._owner = this;
                    this._propertyDescriptors.push(descriptor);
                    this._propertyDescriptorsTable.set(descriptor.name,  descriptor);
                }
                this.addRangeAtPathChangeListener("_ownPropertyDescriptors", this, "_handlePropertyDescriptorsRangeChange");
                this.addRangeAtPathChangeListener("parent.propertyDescriptors", this, "_handlePropertyDescriptorsRangeChange");
            }
        }
    }
},

```

And allow the propertyDescriptors getter to be: 
```javascript
propertyDescriptors: {
        get: function () {
            if (!this._propertyDescriptors) {
                this._propertyDescriptors = [];
            }
            this._preparePropertyDescriptorsCache();
            return this._propertyDescriptors;
        }
    },
```


What do you think?

core/meta/object-descriptor.js
tejaede: Hmm.. Me neither. I would hope that I would not add that unnecessarily, but I cannot see a reason on first inspection. I will remove the slice and run the tests. If all goes well I'll leave it off.

core/meta/property-descriptor.js
tejaede: Correct. This was a part of Jonathan's commit that made DataMapping synchronous whenever possible.  

https://github.com/montagejs/montage/commit/91ce91e309eb68377e3e8d5a4cb4ca6d4d4476b3

I'm not sure why it's showing up as a change in the merge commit. Oddly, the history of that file does not include the merge commit either.  
https://github.com/montagejs/montage/commits/master/core/meta/property-descriptor.js

Perhaps I missed a step when merging over? The PR said that it could be merged without conflict so I did so without performing any other steps. For instance, I did not merge master into the branch before merging the branch back into master. I don't know if that would make a difference. Do you have an idea?

tejaede:
@benoitmarchant  I created a new PR for the changes discussed above. 

https://github.com/montagejs/montage/pull/1922

PR #1892, opened 2017-11-02
Min max converter

marchant:


core/converter/min-max-converter.js
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Converts a number to an integer
+ * @class MinMaxConverter
+ * @extends Converter
+ */
+var MinMaxConverter = exports.MinMaxConverter = Converter.specialize({
+
+    constructor: {
+        value: function MinMaxConverter (min, max) {
```
marchant: Please remove the passing of arguments in constructor, the whole constructor is unecessary

core/converter/min-max-converter.js
```diff
@@ -0,0 +1,66 @@
...
+
+    revert: {
+        value: _clamp
+    }
+    
+
+
+});
+
+// Object.defineProperty(exports, 'singleton', {
```
marchant: Please remove comments

ui/number-field.reel/number-field.js
```diff
@@ -1,40 +1,49 @@
...
    @class module:"montage/ui/native/input-number.reel".InputNumber
    @extends module:montage/ui/text-input.TextInput
  */
 var NumberField = exports.NumberField = TextInput.specialize({
-	hasTemplate: {value: false }
+
+    enterDocument:{
+        value: function() {
+            if(this.min || this.max) {
+                this.converter = new MinMaxConverter(this.min, this.max);
```
marchant: Needs to be modified to not use constructor arguments and instead set .min and .max individually

ui/number-field.reel/number-field.js
```diff
@@ -1,40 +1,49 @@
...
  * Wraps the a &lt;input type="date"> element with binding support for the element's standard attributes.
    @class module:"montage/ui/native/input-number.reel".InputNumber
    @extends module:montage/ui/text-input.TextInput
  */
 var NumberField = exports.NumberField = TextInput.specialize({
-	hasTemplate: {value: false }
+
+    enterDocument:{
+        value: function() {
+            if(this.min || this.max) {
```
marchant: What if either this.min or this.max is not a number/undefined? We should add something like 
this.converter.min = this.min || Infinity;
this.converter.max = this.max || Infinity;

core/converter/min-max-converter.js
```diff
@@ -0,0 +1,36 @@
...
+    return value <= this.min ? this.min : value >= this.max ? this.max : value;
+    
+}
+
+/**
+ * Converts a number to an integer
+ * @class MinMaxConverter
+ * @extends Converter
+ */
+var MinMaxConverter = exports.MinMaxConverter = Converter.specialize({
```
marchant: Let's rename it RangeEnforcer

ui/number-field.reel/number-field.js
```diff
@@ -1,40 +1,50 @@
 /**
-    @module "montage/ui/native/input-number.reel"
-    @requires montage/ui/component
+    @module "montage/ui/number-field.reel"
     @requires montage/ui/text-input
 */
 
-var TextInput = require("ui/text-input").TextInput;
+var TextInput = require("ui/text-input").TextInput,
+    MinMaxConverter = require("core/converter/min-max-converter.js").MinMaxConverter;
```
marchant: Rename to RangeEnforcer

ui/number-field.reel/number-field.js
```diff
@@ -1,40 +1,50 @@
...
+ * Wraps the a &lt;input type="number"> element with binding support for the element's standard attributes.
+   @class module:"montage/ui/number-field.reel".NumberField
    @extends module:montage/ui/text-input.TextInput
  */
 var NumberField = exports.NumberField = TextInput.specialize({
-	hasTemplate: {value: false }
+
+    enterDocument:{
+        value: function() {
+            this.converter = new MinMaxConverter();
```
marchant: RangeEnforcer, we probably don't need to add -Infinity nor Infinity for it to work after all

PR #1899, opened 2017-11-17
Quick Fix for the Translate composer on iOS 11+

thibaultzanini:
The translate composer won’t listen to scroll events after a touch start, Safari raises scroll events all time even if there is nothing to scroll. (Bouncing?).
Need more investigation.


marchant:
This should only impact iOS, LGTM, but let's keep an eye on it

hthetiot:
screen shot 2017-11-20 at 10 45 39 am

Http service fail, not related, could be timeout.

hthetiot:
lgtm may need testing.

PR #1903, opened 2017-11-28
Bugs/data service deserialization

tejaede:


johnnykahalawai:
@tejaede -

Is deserializeSelf async in any other object?  Maybe I'm wrong but it seems like we are hijacking a function to do something we want which is non-standard.

cc: @benoitmarchant

marchant:
montage reviver already expects deserializeSelf to eventually return a thenable object, to deal with objects that can only deserialize themselves asynchronously, I don't think this is hijacking

johnnykahalawai:
OK then!

tejaede:
@johnnykahalawai 

Snippet that Benoit is referring to, if you're curious:
https://github.com/montagejs/montage/blob/master/core/serialization/deserializer/montage-reviver.js#L646-L659


Once I resolve the Travis build I will merge.

PR #1918, opened 2017-12-15
Workaround for mopped mjson

thibaultzanini:
⚠️  experimental
https://github.com/montagejs/mop/pull/81

### DIFF
--- a/core/serialization/deserializer/montage-reviver.js
+++ b/core/serialization/deserializer/montage-reviver.js
@@ -416,6 +416,7 @@ var MontageReviver = exports.MontageReviver = Montage.specialize(/** @lends Mont
         value: function (value, context, label) {
             var self = this,
                 locationId = value.prototype || value.object,
+                isObjectDescriptor = !!(locationId && (locationId.endsWith(".mjson") || locationId.endsWith(".meta"))),
                 module, locationDesc, location, objectName;
 
             if (locationId) {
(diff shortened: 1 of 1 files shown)


tejaede:
@hthetiot @thibaultzanini This works for us, but my understanding from Thibault is that this is a short-term fix that will not be a part of a release. Is that correct? 

So until we have a long-term fix, I plan to build with a branch that consists of [17.1.1-beta](https://github.com/montagejs/montage/tree/17.1.1-beta) plus https://github.com/montagejs/montage/pull/1918/commits/0206f51ee1a550966ea3272410e35eaf7f645bc4 (commit for this PR)

tejaede:
@thibaultzanini @hthetiot That brings me to another question actually. 

Is there any harm in merging this PR into 17.1.1-beta? Put another way, is 17.1.1-beta going to be merged into master in the future?

thibaultzanini:
@tejaede @hthetiot the fix need to be discussed with @marchant first, but the beta tag seems reasonable to me.

hthetiot:
Waiting for @benoitmarchant review to merge and release 17.1.1

tejaede:
I am intermittently unable to load the app in IE11 when I use montage#master due to a race condition (I think).

The following snippets from montage-reviver are the only places that we catch if a module is a string while deserializing. 

https://github.com/montagejs/montage/blob/master/core/serialization/deserializer/montage-reviver.js#L427-L430
```javascript
if (typeof module === "string" &&
   (locationId.endsWith(".mjson") || locationId.endsWith(".meta")) &&
     this._deserializerConstructor.moduleContexts.has(
    (location = context._require.location + locationId)
   ))
```

```javascript
if (isObjectDescriptor && !Promise.is(module) && !module.montageObject) {
    module = context._require.async(locationDesc.moduleId);
}
```

The problem is that these will only catch a non-promise module for `.mjson files`. In IE11, this occurs for JS files as well. In our application for example, it happens when loading `montage/core/meta/module-object-descriptor`

I found that I could resolve it in IE11 by changing `ModuleLoader.getModule()` to only return a promise. However, this change prevents the application from loading in Chrome. 

```javascript
getModule: {
    value: function (moduleId, label) {
        var objectRequires = this._objectRequires,
            _require,
            module;

        if (objectRequires && label in objectRequires) {
            _require = objectRequires[label];
        } else {
            _require = this._require;
        }

        
        module = this.getExports(_require, moduleId);
        

        //Only checks if module exists
        // if (!module) {

        //Forces getModule to return a promise if the module does not have exports yet
        if (!module || typeof module === "string") {
            module = _require.async(moduleId);
        } 
        
        return module;
    }
}
```

tejaede:
Error as it presents itself in Jasmine: 

screen shot 2018-01-09 at 9 50 55 am

thibaultzanini:
Closing new PR: https://github.com/montagejs/montage/pull/1927

hthetiot:
ACK

Issue #1919, opened 2017-12-16
EventTarget changes prevent montage from loading on IE11


tejaede:
`aWindow.EventTarget` is undefined or null at this line
https://github.com/montagejs/montage/blob/master/core/event/event-manager.js#L556

A screenshot of the console error is attached. 

![screen shot 2017-12-15 at 4 21 38 pm](https://user-images.githubusercontent.com/10407370/34065234-c7733164-e1b4-11e7-951b-c48202902c64.png)

PR #1920, opened 2017-12-16
Add detection for EventTarget as it is not available in IE11

tejaede:


marchant:
LGTM!

hthetiot:
https://github.com/montagejs/montage/issues/1919

PR #1922, opened 2017-12-22
Remove redundant arrays & utilize Set constructor

tejaede:
Fixes per the review on https://github.com/montagejs/montage/pull/1889

### DIFF
--- a/core/converter/pipeline-converter.js
+++ b/core/converter/pipeline-converter.js
@@ -1,5 +1,6 @@
 var Converter = require("./converter").Converter,
-    Promise = require("core/promise").Promise;
+    Promise = require("core/promise").Promise,
+    Set = require("collections/set");
 
 /**
  * Converter that chains a series of converters together
@@ -19,9 +20,8 @@ exports.PipelineConverter = Converter.specialize({
 
     _handleConvertersRangeChange: {
         value: function (plus, minus, index) {
-            var all = this.converters ? this._arrayToSet(this.converters) : [],
-                plusSet = this._arrayToSet(plus),
-                minusSet = this._arrayToSet(minus),
+            var plusSet = new Set(plus),
+                minusSet = new Set(minus),
                 converter, i;
 
             for (i = 0; (converter = minus[i]); ++i) {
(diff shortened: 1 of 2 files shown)


core/converter/pipeline-converter.js
```diff
@@ -19,9 +20,8 @@ exports.PipelineConverter = Converter.specialize({
 
     _handleConvertersRangeChange: {
         value: function (plus, minus, index) {
-            var all = this.converters ? this._arrayToSet(this.converters) : [],
-                plusSet = this._arrayToSet(plus),
-                minusSet = this._arrayToSet(minus),
+            var plusSet = new Set(plus),
```
marchant: Converting to a Set preserves the natural order of the Array when iterating on the Set?

core/meta/object-descriptor.js
```diff
@@ -405,34 +393,35 @@ var ObjectDescriptor = exports.ObjectDescriptor = Montage.specialize( /** @lends
...
+            for (i = 0, n = ownDescriptors.length; i < n && isReady; ++i) {
+                descriptor = ownDescriptors[i];
+                isReady = !!(descriptor && descriptor.name);
+            }
+
+            if (isReady) {
+                this._propertyDescriptorsAreCached = true;
+                this._propertyDescriptors = [];
+                this._propertyDescriptorsTable.clear();
                 for (i = 0, n = ownDescriptors.length; i < n && isReady; ++i) {
```
marchant: iReady has to be true here, otherwise that code wouldn't be executed since there's a if(isReady) around, so probably can be removed?

tejaede: Yep. Good eye. That looks like a copy/paste error.

core/converter/pipeline-converter.js
tejaede: Confirmed with tests in collections: 
https://github.com/tejaede/collections/commit/7ff7b959c3c43fee62a2888ec67a022a9bd6b768

tejaede: If you'd like to run them as well, I found I had to manually call `npm link ../../collections` from collections/test in order to run `npm run test:jasmine` successfully.

hthetiot:
lgtm

PR #1926, opened 2018-01-08
object-to-raw-data mapping fixes

tejaede:
PR addresses 3 issues.

**1. Reinstate lazy creation of mapping rules**
Rule creation was switched to be eager in an attempt to remove the raw serialization data from the `ExpressionDataMapping` object. (Lazy creation requires that the raw rules be stored on the object until we are ready to create the "cooked" rules).  
Unfortunately, rule creation requires that `ExpressionDataMapping.objectDescriptor` is fully deserialized. That is not the case when we do rule creation eagerly so back to lazy creation!

**2. Ensure model properties needed in rawData are fetched before mapping**
Given the following constraints: 
- Foo objects have 2 relationships, `bar` and `baz`, neither is prefetched.
- rawFoo.bar is required as a part of a PUT/POST to the remote foo service.

And the following actions.
1. Foo.baz is updated in the application. 
2. User triggers a save.  

Foo.bar is not yet on the model object so we cannot map it to rawFoo.bar in `mapObjectToRawData(object, rawData)`. 

This commit adds logic in `mapObjectToRawDataProperty()` to ensure that any model property required to build the rawData is fetched before the mapping occurs. 

**3. Add support for synchronous reversion of relationships**
Add support for a synchronous converter to map a relationship from object-to-raw-data.

### DIFF
--- a/core/meta/object-descriptor.js
+++ b/core/meta/object-descriptor.js
@@ -404,35 +404,34 @@ var ObjectDescriptor = exports.ObjectDescriptor = Montage.specialize( /** @lends
         }
     },
 
-_preparePropertyDescriptorsCache: {
-    value: function () {
-        var ownDescriptors = this._ownPropertyDescriptors,
-            isReady = true,
-            descriptor, i, n;
-
-        if (!this._propertyDescriptorsAreCached)  {
-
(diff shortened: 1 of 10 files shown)


hthetiot:
Need rebase on master.

marchant:
LGTM

PR #1928, opened 2018-01-17
Commonjs

hthetiot:


hthetiot:
screen shot 2018-01-17 at 2 37 49 pm

hthetiot:
@thibaultzanini  can you see to remove jshint from dep
cause it's plugged to be here
https://github.com/montagejs/montage/blob/commonjs/core/jshint.js
and use dev only for cli tools
cause jshint via require mr have issue due require() bad call in jshint or fact that it's not loading
so we rollback to core/jshint.js custom build for now

hthetiot:
\o/

PR #1934, opened 2018-02-05
 Allow to use the templating mechanism within a repetition

thibaultzanini:


core/serialization/serialization.js
```diff
@@ -97,6 +97,20 @@ var Serialization = Montage.specialize( /** @lends Serialization.prototype # */
...
     },
 
+    removeObject: {
+        value: function (label) {
+            var serializationObject = this.getSerializationObject();
+
+            if (serializationObject && label in serializationObject) {
+                delete serializationObject[label];
+
+                this.initWithObject(serializationObject);
```
marchant: Is it needed to re-do a complete .initWithObject() It's unusual to do that. Performance impact? Side effects on current state?

core/template.js
```diff
@@ -990,6 +990,26 @@ var Template = Montage.specialize( /** @lends Template# */ {
...
+                    if (parameterName === "*" && parameterElement.childElementCount > 0) { // has default
+                        if (argumentElement.childElementCount > 0 ||
+                            !!(argumentElement.firstChild &&
+                                argumentElement.firstChild.data &&
+                                argumentElement.firstChild.data.trim())
+                        ) { // has arguments
+                            var elementIds = this._getChildrenElementIds(parameterElement);
+                            
+                            for (var i = 0, length = elementIds.length; i < length; i++) {
+                                serialization.removeObject(elementIds[0]);
```
marchant: Each call to this.removeObject() will cause a this.initWithObject(serializationObject), sounds inefficient. Maybe we should have instead a removeObjets() that takes an array and do the initWithObject(serializationObject) only once at the end of the loop?

core/template.js
```diff
@@ -990,6 +990,26 @@ var Template = Montage.specialize( /** @lends Template# */ {
...
+                            }
+                        } else {
+                            var range = parameterElement.ownerDocument.createRange();
```
marchant: I think Range can be reused, might be worth looking into that to minimize garbage creation

ui/component.js
```diff
@@ -2327,26 +2327,30 @@ var Component = exports.Component = Target.specialize(/** @lends Component.proto
...
+                                this._element.firstChild.data &&
+                                this._element.firstChild.data.trim())
+                        ) {
                          //We're missing an argument, we're going to check if we have a default
-                             if (parameterElement && parameterElement.childElementCount > 0) {
-                                 range = this._element.ownerDocument.createRange();
-                                 range.selectNodeContents(parameterElement);
-                                 parameterElement.parentNode.replaceChild(range.extractContents(), parameterElement);
+                            if (parameterElement && parameterElement.childElementCount > 0) {
+                                range = this._element.ownerDocument.createRange();
```
marchant: Similar comment about Range, if reuse works, we could create a pool thoughout montage

core/serialization/serialization.js
thibaultzanini: Agreed maybe not the safest way of doing it .

core/template.js
thibaultzanini: I provided a method called `removeObjectsWithLabels` for that purpose

ui/component.js
thibaultzanini: Not the goal of this PR, writing a new ticket

PR #1937, opened 2018-02-07
Adds an application.info

marchant:


hthetiot:
lgtm

core/application.info/sample/ui/main.reel/main.html
```diff
@@ -0,0 +1,35 @@
...
+    <link rel="stylesheet" type="text/css" href="main.css">
+    <script type="text/montage-serialization">
+        {
+            "owner": {
+                "values": {
+                    "element": {"#": "main"}
+                }
+            },
+            "application": {
+            },
```
hthetiot: "applicationPrototype":"test-application" in package id unknown property to me.
 I would prefer in mjson "prototype": "test-application", is that possible?

PR #1941, opened 2018-02-19
The property `allowsMultipleSelection` of the repetition can be bound.

### LINKED ISSUE #1932: allowsMultipleSelection property can't be bind because it's already bound
Can't use a binding on the property allowsMultipleSelection of the repetition because it's already bound internally.

thibaultzanini:
Fix #1932


marchant:
LGTM, merging

PR #1944, opened 2018-02-24
02 18 optimizations

marchant:


hthetiot:
Look good beside addOneRawData doe not return Promise.

marchant:
Should be squared now

PR #1946, opened 2018-03-02
Fix translate composer for chrome.

thibaultzanini:
Make sure the values of clientX and clientY are integers.


thibaultzanini:
Not a bug need to set the property `allowFloats` to `true`

PR #1948, opened 2018-03-06
Fixed Control converter issue from user input

marchant:
- fixes a control bug with converter in revert so it works while editing
- adds a spec converter use cases for text field
- code re-factoring of rotation composer,

### DIFF
--- a/test/all.js
+++ b/test/all.js
@@ -97,6 +97,7 @@ module.exports = require("montage-testing").run(require, [
     {name: "spec/ui/slot-spec", node: false},
     {name: "spec/ui/substitution-spec", node: false},
     {name: "spec/ui/text-input-spec", node: false},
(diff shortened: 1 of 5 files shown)


test/all.js
```diff
@@ -97,6 +97,7 @@ module.exports = require("montage-testing").run(require, [
     {name: "spec/ui/slot-spec", node: false},
     {name: "spec/ui/substitution-spec", node: false},
     {name: "spec/ui/text-input-spec", node: false},
+    {name: "spec/ui/control/text-field-spec", node: false}
```
hthetiot: Missing comma

marchant:
Thanks @hthetiot pushed a fix

hthetiot:
sry one more lint @benoitmarchant 
> ui/control.js: line 343, col 33, Misleading line break before '?'; readers may interpret this as an expression boundary.

PR #1951, opened 2018-03-14
Html Fragment

thibaultzanini:
Component that allows to parse html strings.


hthetiot:
Lgtm

ui/html-fragment.reel/html-fragment.js
```diff
@@ -0,0 +1,223 @@
...
+                    }
+                }                
+            }
+
+            return parent;
+        }
+    },
+
+    sanitizeHtml: {
+        value: function (html, options) {
```
marchant: remove options and pass allowedTags and allowedAttributes instead

ui/html-fragment.reel/html-fragment.js
```diff
@@ -0,0 +1,223 @@
...
+    });
+
+/**
+ *
+ * @class HtmlFragment
+ * @extends Component
+ */
+var HtmlFragment = exports.HtmlFragment = Component.specialize(/** @lends HtmlFragment# */ {
+
+    initWithOptions: {
```
marchant: remove initWithOptions and .options

ui/html-fragment.reel/html-fragment.js
```diff
@@ -0,0 +1,223 @@
...
+
+    allowedTags: {
+        value: null
+    },
+
+    allowedAttributes: {
+        value: null
+    },
+
+    _getSanitizerOptions: {
```
marchant: replace by 2 getters, one for allowedTags and one for allowedAttributes

ui/html-fragment.reel/html-fragment.js
```diff
@@ -0,0 +1,223 @@
...
+                this.element.innerHTML = '';
+            }
+
+            this.needsSanitizeHtml = false;
+        }
+    }
+  
+}, {
+    
+    DefaultSanitizerOptions: {
```
marchant: might be worth exposing on prototype as well? -> defaultAllowedTags and defaultAllowedAttributes

ui/html-fragment.reel/html-fragment.js
thibaultzanini: will do

ui/html-fragment.reel/html-fragment.js
thibaultzanini: will do

ui/html-fragment.reel/html-fragment.js
thibaultzanini: @marchant Having a function for getting the options could help us for future features, such as filtering css rules, replacing tag names...
 Moreover `_sanitizeHtml` is an internal and private method.

thibaultzanini:
@marchant Having a function for getting the options could help us for future features, such as filtering css rules, replacing tag names...
 Moreover `_sanitizeHtml` is an internal and private method.

marchant:
yes it's internal, but we're creating an intermediary option object every time just to move the different aspects, internally. That won't change conceptually for future/more aspects, it will mean more of the same pattern. Not a huge deal as there will rarely be a ton of html-fragment instances, but garbage in, garbage out!

thibaultzanini:
@marchant done

ui/html-fragment.reel/html-fragment.js
```diff
@@ -0,0 +1,217 @@
...
+                                    if (attributeName.startsWith('data-') &&
+                                        allowedAttributesForTag.indexOf('data-*') > -1
+                                    ) {
+                                        shouldRemoveAttribute = false;
+                                    }
+                                }
+
+                                if (shouldRemoveAttribute) {
+                                    delegateResponse = this.callDelegateMethod(
+                                        'htmlFragmentWillRemoveAttributeFromNode',
```
marchant: let's rename to htmlFragmentWillRemoveNodeAttribute (this, node, attribute)

ui/html-fragment.reel/html-fragment.js
```diff
@@ -0,0 +1,217 @@
...
+                                    }
+
+                                    if (shouldRemoveAttribute) {
+                                        child.removeAttribute(attributeName);
+                                        ll--;
+                                        ii--;
+                                    }
+                                } else {
+                                    delegateResponse = this.callDelegateMethod(
+                                        'htmlFragmentWillUseAttributeValueForAttributeNameFromNode',
```
marchant: let's rename to htmlFragmentWillUseValueForNodeAttribute (this, value, node, attribute)

ui/html-fragment.reel/html-fragment.js
```diff
@@ -0,0 +1,217 @@
...
+        }
+    },
+
+    _sanitizeHtml: {
+        value: function (html, allowedTags, allowedAttributes) {
+            var doc;
+
+            if (window.DOMParser) {
+                try {
+                    doc = new DOMParser().parseFromString(html, "text/html");
```
marchant: We may not need to create a DOMParser every time but instead share one?

ui/html-fragment.reel/html-fragment.js
```diff
@@ -0,0 +1,217 @@
...
+        value: function () {
+            if (this.value && this.needsSanitizeHtml) {
+                var doc = this._sanitizeHtml(
+                    this.value,
+                    this.allowedTags || defaultOptions.allowedTags,
+                    this.allowedAttributes || defaultOptions.allowedAttributes
+                );
+
+                if (doc) {
+                    var range = doc.createRange();
```
marchant: Same here, I think we can share a Range object among instances since it's not recursive and Range's state is reset here everytime it's used

PR #1952, opened 2018-03-16
Adds method to reset data object to last snapshot

johnnykahalawai:
@marchant - per our discussion, I added the capability to reset a single data object.

### DIFF
--- a/data/service/raw-data-service.js
+++ b/data/service/raw-data-service.js
@@ -284,6 +284,24 @@ exports.RawDataService = DataService.specialize(/** @lends RawDataService.protot
     },
 
     /**
+     *
+     * Resets the object to its last known state.
+     *
+     * @method
+     * @argument {Object} object   - The object to reset.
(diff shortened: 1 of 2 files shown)


johnnykahalawai:
Useful for use in combination with a form e.g. when the user cancels the form the data object can be mapped back to its last known state.

data/service/raw-data-service.js
```diff
@@ -808,6 +805,12 @@ exports.RawDataService = DataService.specialize(/** @lends RawDataService.protot
      *                             [addRawData()]{@link RawDataService#addRawData}
      *                             call that invoked this method.
      */
+     mapRawDataToObject: {
```
marchant: doesn't that break backward compatibility or are we now in position to do so?

johnnykahalawai: will fix - the code comments were not lined up so I put it over the right function.  There was a merge conflict and I removed the wrong block.

PR #1953, opened 2018-03-19
Fix succession sample and fix an issue with the button.

thibaultzanini:
Allow to use the text content of the original element.


marchant:
LGTM

Issue #1961, opened 2018-04-04
Deserializer deserializeObject fail on nodeJS


hthetiot:
Current Montage Reviver issue that works on client but not nodeJS is following:

```
var mainService;
function getMainService() {

    // Once only
    if (mainService) {
        return Promise.resolve(mainService);
    }

    return getMontageRequire().then(function (mr) {
        return mr.async('montage/core/serialization/deserializer/montage-deserializer').then(function (module) {
            var Deserializer = module.MontageDeserializer;
            return mr.async('data/main.mjson').then(function (descriptor) {
                var deserializer = new Deserializer().init(descriptor, mr);
                return deserializer.deserializeObject();
            }); 
        });
    });
};

// ok on client fail on node with error bellow
getMainService().then(function (mainService) {
    console.log(typeof mainService.fetchData == 'function', mainService);
});
```

cc @thomas @benoit @johnnykahalawai 

```
Unhandled rejection TypeError: Cannot read property 'defineDeserializationUnit' of undefined
    at Module.eval (file:///me/node_modules/montage/core/serialization/bindings.js:118:13)
    at Module.module.factory (/me/node_modules/montage/node_modules/mr/node.js:72:32)
    at Module.module.factory (/me/node_modules/montage/node_modules/mr/require.js:1259:52)
    at getExports (/me/node_modules/montage/node_modules/mr/require.js:632:38)
    at require (/me/node_modules/montage/node_modules/mr/require.js:707:24)
    at Module.eval (file:///me/node_modules/montage/core/core.js:1730:1)
    at Module.module.factory (/me/node_modules/montage/node_modules/mr/node.js:72:32)
    at Module.module.factory (/me/node_modules/montage/node_modules/mr/require.js:1259:52)
    at getExports (/me/node_modules/montage/node_modules/mr/require.js:632:38)
    at require (/me/node_modules/montage/node_modules/mr/require.js:707:24)
    at Module.eval (file:///me/node_modules/montage/core/serialization/deserializer/montage-deserializer.js:1:49)
    at Module.module.factory (/me/node_modules/montage/node_modules/mr/node.js:72:32)
    at Module.module.factory (/me/node_modules/montage/node_modules/mr/require.js:1259:52)
    at getExports (/me/node_modules/montage/node_modules/mr/require.js:632:38)
    at Module.require [as mappingRequire] (/me/node_modules/montage/node_modules/mr/require.js:707:24)
[...]

PR #1962, opened 2018-04-04
Fix NodeJS Serialization support

hthetiot:
Revert https://github.com/montagejs/montage/commit/476cd558d
See #1961


hthetiot:
Look good to me.

PR #1963, opened 2018-04-05
Cache single resolved Promise to use in place of Promise.resolve

### LINKED ISSUE #1875: Cache single resolved Promise to use in place of Promise.resolve()
Throughout montage (see template.js, document-resources.js for examples), Promise.resolve() is used to allow methods that return promises to maintain a clean API in the event that they do not have other async work to do. 

From template.js: 
```javascript
loadScripts: {
    value: function (targetDocument) {
        var scripts = this.getScripts(),
            ii = scripts.length;

        if (ii) {
            var promises = [];

            for (var i = 0; i < ii; i++) {
                promises.push(
                    this.loadScript(scripts[i], targetDocument)
                );
            }

            return Promise.all(promises);
        }

        return Promise.resolve();
    }
},
```

Each Promise.resolve() call creates a new (albeit short lived) Promise object. Instead, I propose caching a single resolved promise on the class and returning that whenever necessary. For example, Template could have the following property:

```javascript
resolvedPromise: {
    get: function () {
        if (!exports.Template._resolvedPromise) {
            exports.Template._resolvedPromise = Promise.resolve();
        }
        return exports.DataService._resolvedPromise;
    }
},
```

And then the return statement for the loadScripts function() above becomes: 
```
return this.resolvedPromise;
```

thibaultzanini:
Fix #1875

### DIFF
--- a/core/template.js
+++ b/core/template.js
@@ -1384,6 +1384,15 @@ var TemplateResources = Montage.specialize( /** @lends TemplateResources# */ {
         }
     },
 
+    resolvedPromise: {
+        get: function () {
+            if (!TemplateResources._resolvedPromise) {
+                TemplateResources._resolvedPromise = Promise.resolve();
+            }
+            return TemplateResources._resolvedPromise;
+        }
+    },
+
     initWithTemplate: {
         value: function (template) {
             this.template = template;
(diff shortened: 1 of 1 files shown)


hthetiot:
lgtm

PR #1966, opened 2018-04-11
Remove deprecated custom property descriptor `distinct`

### LINKED ISSUE #1556: Remove deprecated custom property descriptor `distinct`
`distinct` is deprecated per:
https://github.com/montagejs/montage/blob/master/core/core.js#L288

Remove related code in `core.js`, [related specs](https://github.com/montagejs/montage/blob/master/test/core/core-spec.js#L640) & grep code to remove existing `distinct` on properties.

thibaultzanini:
Fix #1556


hthetiot:
lgtm

PR #1972, opened 2018-04-21
Add custom serialization on DataQuery

tejaede:
Add serializeSelf and deserializeSelf to DataQuery to allow DataQuery#type to be passed as a module instead of a full object descriptor.


tejaede:
@cdebost Can you review and test this with MontageStudio, if possible?

marchant:
👍🏻

PR #1976, opened 2018-05-07
Data operation evolution

marchant:


marchant:
WIP, not ready to merge

data/model/data-query.js
```diff
@@ -96,6 +96,15 @@ exports.DataQuery = Montage.specialize(/** @lends DataQuery.prototype */ {
         value: undefined
     },
 
+    /**
+     * A property used to give a meaningful name to an operation
+     *
+     * @type {string}
+     */
+    name: {
```
tejaede: Per the comment, is this meant to be on DataOperation instead of DataQuery? 

Or if it's meant to be DataQuery, can you give an example of it's usage?

data/service/data-operation.js
```diff
@@ -147,14 +341,65 @@ exports.DataOperation = Montage.specialize(/** @lends DataOperation.prototype */
...
-            DELETE: {isDelete: true}
+            Create: {isCreate: true},
+            CreateFailed: {isCreate: true},
+            CreateCompleted: {isCreate: true},
+            //Additional
+            Copy: {isCreate: true},
+            CopyFailed: {isCreate: true},
+            CopyCompleted: {isCreate: true},
+            /* Read is the first operation that mnodels a query */
+            Read: {isRead: true},
```
marchant: Do we need a "ReadProperty" and parallel set of accompanying operations to model when we fetch individual properties of objects?

tejaede: That an operation is meant to read a property strikes me as a characteristic of a "Read" operation rather than a separate kind of operation altogether. If the flow for a read goes: 
1. Is this a read? Check `anOperation.type`. 
2. What are we reading? Check `anOperation.criteria`. 

Then we should store the information about "is this for a property" in the criteria, not the type. 

Or consider an `Update` operation.  `Update` doesn't say in and of itself whether a single property or every property is modified. The specifics about which properties are modified are passed with the data and criteria. It is not several `UpdateProperty` operations.

hthetiot:
```
data/service/data-operation.js: line 62, col 5, Expected '}' to match '{' from line 10 and instead saw 'name'.
data/service/data-operation.js: line 62, col 9, Expected ')' and instead saw ':'.
```

Happy Travis.

data/service/data-operation.js
marchant: 👍🏻

marchant: TODO: It's missing authentication related API

marchant:
No jshint issues locally, merging

PR #1978, opened 2018-05-14
Merge master in commonjs

hthetiot:


hthetiot:
mop failed but that expected

PR #1979, opened 2018-05-14
Backend-for-Montage support

tejaede:
Refactor and build out Montage Data to enable "Sentinel" project 

This PR encapsulates the following functionality: 

1. **RawDataWorker** Receives DataOperation objects and lazily initializes RawDataServices designed to handle those DataOperations.

2. **DataService Mapping** Move responsibility for mapping of raw data to cooked objects and vice versa into DataService. This allows a single DataService to handle all of the data for an application without maintaining the full RawDataService tree. It follows that RawDataServices (and the responsibility to fetch data from the origin) can move off the client.

3. **DataOperation refactor** Build out DataOperation API to normalize the way data travels up and down the stack.


tejaede:
@marchant 

The values in `DataOperation.Type` aren't distinguishable after serialization. At the moment, it is possible to narrow them down by their CRUD action using `isRead`, `isCreate` etc, but it is not possible to distinguish a Read from a ReadCompleted from a ReadFailed. In order for it to be possible, the class needs another property in addition to (or in place of) the is<action> flags.  For example, an id property that allowed for something like `operationA.type.id === operationB.type.id`

Further, in order for them to be identifiable with equality in memory, it would need to be a proper class instead of a POJO.  Then the check is as simple as `operationA.type === operationB.type`. This is the best option for efficiencies sake. With the number of operations we will be processing at each level, only 20 operation type objects floating around is a great improvement from potentially thousands of POJOs.

PR #1981, opened 2018-05-31
Drag Manager

thibaultzanini:
- [x] basic drag support
- [x] basic drop support
- [x] outer/inner drop support
- [x] support touch devices
- [x] contain dragging operation within rectangle bound
- [x] drop files support
- [x] move, copy, link operation support
- [x] scroll while dragging
- [x] drag-manager.info
- [x] specs
- [x] comments


ui/component.js
```diff
@@ -2904,6 +2970,205 @@ var Component = exports.Component = Target.specialize(/** @lends Component.proto
...
+        }
+    },
+
+    /**
+     * Called when the drag operation start,
+     * allowing the receivers to agree to or refuse the drag operation.
+     * By default all receivers will not accept any drag operation.
+     * The method `draggingStarted` should return a boolean value.
+     */
+    _draggingStarted: {
```
marchant: @marchant Needs to find a matching event name, to tell a potential destination that a drag started

core/drag/drag-event.js
```diff
@@ -0,0 +1,343 @@
...
+
+        hasData: {
+            value: function (key) {
+                return this._data.has(key);
+            }
+        },
+
+        setData: {
+            value: function (key, value) {
+                return this._data.set(key, value);
```
thibaultzanini: Add specified mime type for objects `application/object`

add sugar method `draggedObject`

core/drag/drag-event.js
```diff
@@ -0,0 +1,343 @@
...
+
+        items: {
+            value: null
+        },
+
+        types: {
+            value: null
+        },
+
+        dragTarget: {
```
thibaultzanini: rename it to `dropTarget`

thibaultzanini: remove it

core/drag/drag-event.js
```diff
@@ -0,0 +1,343 @@
...
+
+        candidateDropTargets: {
+            get: function () {
+                return this._candidateDropTargets ||
+                    (this._candidateDropTargets = new Set());
+            }
+        },
+
+        dragabblePlaceholderStrategy: {
+            value: 'hidden'
```
thibaultzanini: `showPlaceholder` -> boolean

core/drag/drag-event.js
```diff
@@ -0,0 +1,343 @@
...
+
+        setData: {
+            value: function (key, value) {
+                return this._data.set(key, value);
+            }
+        },
+
+        setDragImage: {
+            value: function (img, xOffset, yOffset) {
+                if (!this._dragImage) {
```
thibaultzanini: store and use xOffset, yOffset args

core/drag/drag-manager.js
```diff
@@ -0,0 +1,1287 @@
...
+        value: null
+    },
+
+    _droppables: {
+        get: function () {
+            return this.__droppables || (this.__droppables = []);
+        }
+    },
+
+    _dropDestination: {
```
thibaultzanini: add this property to the dragging operation context

core/drag/drag-manager.js
```diff
@@ -0,0 +1,1287 @@
...
+
+            return this.__translateComposer;
+        }
+    },
+
+    _draggingOperationContext: {
+        value: null
+    },
+
+    _isDragging: {
```
thibaultzanini: add the properties bellow to the context

core/drag/drag-event.js
```diff
@@ -0,0 +1,343 @@
...
+        dragTarget: {
+            value: null
+        },
+
+        _candidateDropTargets: {
+            value: null,
+            enumerable: false
+        },
+
+        candidateDropTargets: {
```
thibaultzanini: rename `dropTargetCandidates`

core/drag/drag-event.js
```diff
@@ -0,0 +1,371 @@
+var MutableEvent = require("../event/mutable-event").MutableEvent,
+    Montage = require("../core").Montage;
+
+// XXX Does not presently function server-side
+if (typeof window !== "undefined") {
+
+    var OBJECT_MINE_TYPE = 'application/object';
```
marchant: Should be  OBJECT_MIME_TYPE (M)

core/drag/drag-event.js
```diff
@@ -0,0 +1,371 @@
...
+            enumerable: false,
+        },
+
+        _dragEffect: {
+            value: null
+        },
+
+        dragEffect: {
+            set: function (effect) {
+                if (DataTransfer.ALLOWED_EFFECTS.indexOf(effect) > -1) {
```
marchant: We're repeating this: if (DataTransfer.ALLOWED_EFFECTS.indexOf(effect) > -1) quite a bit. I think it would make sense to create a method on DataTransfer like isAllowedEffect(effect) instead. And internally it might be faster to keep it as a table or object with properties rather than an array? For small amount of objects involved I'm not sure, but at least it would become an implementation detail.

core/drag/drag-event.js
```diff
@@ -0,0 +1,371 @@
...
+
+            All: {
+                value: "all"
+            },
+
+            _ALLOWED_EFFECTS: {
+                value: null
+            },
+
+            ALLOWED_EFFECTS: {
```
marchant: Not sure why this is all-caps?

core/drag/drag-manager.js
```diff
@@ -0,0 +1,1314 @@
...
+     * @private
+     * @function
+     * @param {DraggingOperationContext} draggingOperationContext - current dragging 
+     * operation info object
+     * @description Dispatch Drag Leave Event
+     */
+    _dispatchDragLeave: {
+        value: function (draggingOperationContext) {
+            if (draggingOperationContext.currentDropTarget) {
+                draggingOperationContext.currentDropTarget.classList.remove("drag-over");
```
marchant: classList.remove( String [, String [, ...]] ), no need for 2 calls

core/drag/drag-manager.js
```diff
@@ -0,0 +1,1314 @@
...
+    },
+
+    /**
+     * @private
+     * @function
+     * @description add translate listeners.
+     */
+    _addTranslateListeners: {
+        value: function () {
+            this._translateComposer.addEventListener('translate', this);
```
marchant: shouldn't there be a third argument for capture/bubble? Remember in the past Firefox barfing on that

core/drag/drag-manager.js
```diff
@@ -0,0 +1,1314 @@
...
+                        draggingOperationContext.positionX,
+                        draggingOperationContext.positionY
+                    );
+
+                    if (droppable && 
+                        !draggingOperationContext.dropTargetCandidates.has(
+                            droppable
+                        )
+                    ) {
+                        droppable.classList.add('invalid-drop-target');
```
marchant: The DOM is not supposed to be touched in willDraw. Do we anticipate this logic to potentially trigger some components to be added to the current draw loop? If not, it might be better to move it to the beginning of draw()

core/drag/drag-manager.js
```diff
@@ -0,0 +1,1314 @@
...
+                    if (draggingOperationContext.dataTransfer.dragImageYOffset > draggedImageBoundingRect.height) {
+                        top -= draggedImageBoundingRect.height;
+                    } else {
+                        top -= draggingOperationContext.dataTransfer.dragImageYOffset;
+                    }
+                } else {
+                    top = draggedImageBoundingRect.top;
+                }
+
+                draggedImage.style.top = top + PX;
```
marchant: better to do top += PX; and then 
draggedImage.style.top = top;

core/drag/drag-manager.js
```diff
@@ -0,0 +1,1314 @@
...
+
+                if (draggingOperationContext.dragEffect === "move") {
+                    var draggableElement = draggingOperationContext.draggable.element;
+                    this._oldDraggableDisplayStyle = draggableElement.style.display;
+                    draggableElement.style.display = 'none'; 
+
+                    if (draggingOperationContext.showPlaceholder) {
+                        var placeholderElement = document.createElement('div');
+                        placeholderElement.style.width = (
+                            draggedImageBoundingRect.width + PX
```
marchant: concatenation of string creates garbage that way, += better

core/drag/drag-manager.js
```diff
@@ -0,0 +1,1314 @@
...
+                        );
+                        placeholderElement.style.height = (
+                            draggedImageBoundingRect.height + PX
```
marchant: concatenation of string creates garbage that way, += better

core/drag/drag-manager.js
```diff
@@ -0,0 +1,1314 @@
...
+                } else {
+                    this._draggingOperationContext = null;
+                }
+            }
+
+            this._removeDraggablePlaceholderIfNeeded();
+        }
+    },
+
+    _setUpDraggedImageIfNeeded: {
```
marchant: lets call this _drawDraggedImageIfNeeded instead, that way it's clear this is is a piece of the draw

core/drag/drag-event.js
thibaultzanini: constants

thibaultzanini: not anymore actually they became maps

core/drag/drag-manager.js
thibaultzanini: The DOM is not touched here. the droppable variable is a pointer to a component so it will be automatically added/removed here:
https://github.com/montagejs/montage/blob/master/ui/component.js#L3329

thibaultzanini:
@marchant Changes done

core/drag/drag-event.js
```diff
@@ -86,7 +82,7 @@ if (typeof window !== "undefined") {
 
         effectAllowed: {
```
marchant: I think this should be "allowedEffect" instead from an english stand point

marchant:
@thibaultzanini Just one last change needed on drag-event, I think effectAllowed should be allowedEffect instead, just from an english stand point

thibaultzanini:
@marchant https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/effectAllowed

PR #1984, opened 2018-06-15
Add specs for authorization API

tejaede:
This PR is intended to add Authorization specs. 

Adding those specs uncovered 2 opportunities for functional changes that are also included.

#### 1. Add ability to clear authorization for a dataService (e.g. on expiration or log out) 
 - Accomplished with new method `Authorization.clearAuthorizationForService()`

####  2. Add singleton AuthorizationManagerPanel to AuthorizationManager class
- AuthorizationManager itself is a singleton and was responsible for creating the manager panel, which therefore was a singleton as well. This PR simply moves the manager panel creation to the AuthorizationManagerPanel class to mimic the creation of other Montage singletons e.g. Environment, AuthorizationManager


test/run.html
```diff
@@ -10,5 +10,6 @@
 	<script type="text/javascript" src="../node_modules/jasmine-core/lib/jasmine-core/jasmine-html.js"></script>
 	<script type="text/javascript" src="../test/run-browser.js" ></script>
 	<script type="text/javascript" src="../montage.js" data-module="./all"></script>
+	<!-- <script type="text/javascript" src="../montage.js" data-module="./auth-specs.js"></script> -->
```
marchant: Any reason to leave that comment?

ui/authorization-manager-panel.reel/authorization-manager-panel.js
```diff
@@ -103,6 +124,18 @@ exports.AuthorizationManagerPanel = Component.specialize({
         }
     }
 
+}, {
+
+    instance: {
```
marchant: "instance" is technically correct, but not a great API name

data/service/authorization-manager.js
```diff
@@ -163,24 +179,31 @@ var AuthorizationManager = Montage.specialize(/** @lends AuthorizationManager.pr
...
 
     _authorizeProviderWithManagerPanel: {
         value: function (provider) {
             var self = this,
-                managerPanel;
-            self._pendingServicesCount++;
+                managerPanel,
+                result = null;
+            
+            self._pendingServices.add(provider.identifier);
```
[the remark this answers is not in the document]
tejaede: Fixed: https://github.com/montagejs/montage/commit/aeb0e486c3006a42d095f76ad7038d17a39c6de9

ui/authorization-manager-panel.reel/authorization-manager-panel.js
tejaede: https://github.com/montagejs/montage/commit/aeb0e486c3006a42d095f76ad7038d17a39c6de9

test/run.html
tejaede: Nope! Removed. https://github.com/montagejs/montage/pull/1984/commits/39e7b828333474a4fc664667a8c6e91e88fac7a8

data/service/authorization-manager.js
```diff
@@ -259,38 +282,31 @@ var AuthorizationManager = Montage.specialize(/** @lends AuthorizationManager.pr
...
             var self = this,
                 authorizationPromises = [];
 
-
             if (dataService.authorizationPolicy === AuthorizationPolicy.NONE) {
                 return Promise.resolve(null);
             } else {
-
-                //[TJ] This will only work for data services with a single authorization-service
+                //[TJ] This will only work for data services with a single authorization-service
```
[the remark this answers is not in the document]
tejaede: @marchant 
The comment has a kernel of truth, but isn't wholly accurate. I removed the comment in code to avoid further confusion (https://github.com/montagejs/montage/commit/063f5465fdb34c8ef259f842559db008810de0b2) and will clarify below....

`AuthorizationManager` can handle a `DataService` with multiple authorization services, but only the case where the `DataService` requires *all* of the authorization services to resolve before it considers itself authorized. It does not support the user choosing only one of the provided authorization services. This is a case we must handle. Further, I think that case was the original intention. 

That said, this PR contains useful additions that I do not want to delay for the sake of the multiple auth-service issue. I would prefer to move that to its own issue/PR.

PR #1988, opened 2018-06-27
Fix #1971

### LINKED ISSUE #1971: Default Button Label appears in child classes of Button
The default label in Button breaks backwards compatibility in classes that specialize Button. For example, take the following:
```javascript
var Button = require("montage/ui/button.reel").Button;

exports.ActionButton = Button.specialize(/** @lends ActionButton.prototype */ {
```

With [this commit](https://github.com/montagejs/montage/commit/14c94ebc202b93b091262ee6ed4eb16f1f865661#diff-25ae8e394180e7b19d82a350feef5a1e) "Button" is shoehorned into the element as seen below.

screen shot 2018-04-19 at 1 42 41 pm
screen shot 2018-04-19 at 1 43 27 pm

A quick fix on our end is to define empty `label` and `defaultLabel` properties in the Child class
```javascript
var Button = require("montage/ui/button.reel").Button;

exports.ActionButton = Button.specialize(/** @lends ActionButton.prototype */ {

    defaultLabel: {
        value: ""
    },

    label: {
        value: ""
    },
```

However, the label is a breaking change if it is released and changes are not made in the implementing application.

For context, the use case for specializing button is two-fold: 
1.  Custom Styling 
2. Act as link if a url is provided

thibaultzanini:


tejaede:
To take a step back, I'm not sure I understand the value of a default label for button. The component does not need a label to work and I think the total amount of work for the developer is greater if button has a default.

If you boil down the possible options: 
1. No Label 
2. Label is the default
3. Label is NOT the default

"Label is NOT the default" is going to require the same amount of work regardless so set that option aside.

For the others.... 
Without a default, "No Label" requires no code while "Label is the default" does.
With a default, "Label is the default" requires no code while "No Label"  does.

In my eyes, "No Label" is significantly more likely than using the default so we should defer to the  implementation that minimizes the work in that case. 


Anecdotal evidence is here: https://github.com/thibaultzanini/montage/commit/14c94ebc202b93b091262ee6ed4eb16f1f865661#diff-b15e34a728b9cfd69f2e35b233b4d780

thibaultzanini:
@tejaede I agree it's more work for the developper. @marchant asked me to do that change for the studio and now I believe the studio should set this default value. FYI @cdebost

PR #1992, opened 2018-07-18
Fix minor montage-data bugs

tejaede:
This PR fixes 2 issues.

1. Add `isUniquing` to `DataService.deserializeSelf()`

2. Assign an inverseProperty before assigning the requested relationship. 
For example, say you have a relationship from Foo to Bar where: 
- `Foo.bars` is an array of `Bar` objects 
- Each item in `Foo.bars` is assigned the`Foo` as `Bar.foo`

As in the following mapping
```javascript
{
    "root": {
        "prototype": "montage/data/service/expression-data-mapping",
        "values": {
            "rawDataPrimaryKeys":["id"],
            "objectDescriptor": {"@": "FooDescriptor"},
            "objectMapping": {
                "rules": {
                    "bars": {
                        "<-": "{id: id}",
                        "converter": {"@": "barsConverter"},
                        "inversePropertyName": "foo"
                    }
                }
            }
        }
    },

    "FooDescriptor": {
        "object": "data/descriptors/foo.mjson"
    },

    "barsConverter": {
        "prototype": "montage/data/converter/raw-property-value-to-object-converter",
        "values": {
            "convertExpression": "id == id"
        }
    }
}
```

The mapping establishes the contract that when `Foo.bars` is fetched, it is only returned once each item in `Foo.bars` has `item.foo`. However, the promise returned by `DataTrigger.updateObjectProperty()` resolves as soon as the value is set. So if we assign `Foo.bars` before assigning `bar.foo`, the following occurs:
```javascript
mainService.updateObjectProperties(aFoo, "bars").then(function () {
          // items in aFoo.bars do not yet have a foo property
});
```

This PR ensures that `bar.foo` is assigned before `Foo.bars`

### DIFF
--- a/data/service/data-service.js
+++ b/data/service/data-service.js
@@ -87,6 +87,11 @@ exports.DataService = Montage.specialize(/** @lends DataService.prototype */ {
                 this.delegate = value;
             }
 
+            value = deserializer.getProperty("isUniquing");
+            if (value !== undefined) {
+                this.isUniquing = value;
+            }
+            
             return result;
         }
     },
--- a/data/service/data-service.js
+++ b/data/service/data-service.js
@@ -1295,8 +1295,8 @@ exports.DataService = Montage.specialize(/** @lends DataService.prototype */ {
 
             // Check if property is included in debugProperties. Intended for debugging
             if (debug) {
-                console.log("DataService.fetchObjectProperty", object, propertyName);
-                console.log("To debug DataService.mapRawDataToObjectProperty for " + propertyName + ", place a breakpoint at this line in your browser's developer tools.");
+                console.debug("DataService.fetchObjectProperty", object, propertyName);
+                console.debug("To debug ExpressionDataMapping.mapRawDataToObjectProperty for " + propertyName + ", set a breakpoint here.");
             }
 
             return  useDelegate ?                       this.fetchRawObjectProperty(object, propertyName) :
--- a/data/service/data-service.js
+++ b/data/service/data-service.js
@@ -56,7 +56,7 @@ exports.DataService = Montage.specialize(/** @lends DataService.prototype */ {
     deserializeSelf: {
         value:function (deserializer) {
             var self = this,
-                result = null,
+                result = this,
                 value;
 
             value = deserializer.getProperty("childServices");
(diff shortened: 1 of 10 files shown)


marchant:
@tejaede, looks good, but I have a feeling inverseProperty[Name] belongs more to the Object/PropertyDescriptor level than the mapping, where it would still be available.

tejaede:
@marchant True. PR is updated accordingly. 

https://github.com/montagejs/montage/pull/1992/commits/104e491895abc402317f9c5d1a26848848defbac does 3 things: 
1. Adds `PropertyDescriptor.inversePropertyName`
2. Deprecates `MappingRule.inversePropertyName`
3. Adds support for `PropertyDescriptor.inversePropertyName` in `ExpressionDataMapping`

data/service/data-service.js
```diff
@@ -1283,7 +1289,20 @@ exports.DataService = Montage.specialize(/** @lends DataService.prototype */ {
...
+                trace = debug || exports.DataService.traceProperties.has(propertyName);
+
+
+            // Check if property is included in the traceProperties or debugProperties 
+            // collections. Intended for debugging
+            if (trace) {
+                console.log("DataService.fetchObjectProperty", object, propertyName);
+            }
+            
+            if (debug) {
```
marchant: Event with a debug conditional, we can't have a debugger; statement in production code as it can derail runtime optimizations

data/service/data-stream.js
```diff
@@ -357,6 +357,16 @@ exports.DataStream = DataProvider.specialize(/** @lends DataStream.prototype */
             stream.query = selector;
             return stream;
         }
+    },
```
marchant: Why not have withTypeOrSelector states a deprecation message (only once) and then call the new withTypeOrQuery, no need to have twice the code

data/service/data-service.js
tejaede: @marchant 

As a general rule I agree. I saw the pattern used in MontageReviver ([line 351](https://github.com/montagejs/montage/blob/master/core/serialization/deserializer/montage-reviver.js#L351)) and realized it would be useful if applied here. In this case, I found myself regularly writing `if (propertyName == 'aProperty')` in data-service and data-mapping. This change allows debugging a property fetch without modifying any montage code. 

In the end, I agree it is an anti-pattern, but the functionality of the pattern is useful and I do not want to lose it if possible. Do you have any ideas on how it might be accomplished in another way?

data/service/data-stream.js
tejaede: Good point! Done. 

https://github.com/montagejs/montage/pull/1992/commits/b23207fdb2a83b2b41e169dbd4f54f098250a4ae

tejaede:
@marchant Should be good! The `DataService.debugProperties` code uses `console.debug` as of https://github.com/montagejs/montage/pull/1992/commits/25a66be2f9a987adb8c0c77e719f8c0e96b03eaa

PR #1999, opened 2018-11-14
Force localizer to use mr as default require

tejaede:
localizer is designed to run with mr, however, it is
possible for global.require to be overridden by another
module when localizer is loaded.


marchant:
@tejaede what use case have you found where global.require is overridden by another
module, causing trouble when localizer is loaded?

tejaede:
@marchant When used with cordova.

PR #2001, opened 2018-12-11
Adds setters to MutableEvent's touch events

tejaede:


tejaede:
@marchant If I recall, we are agreed that this PR is safe and can be merged into master. If that is still the case, will you please merge?

core/event/mutable-event.js
```diff
@@ -236,6 +236,13 @@ if (typeof window !== "undefined") {
...
             get: function () {
                 return this._event.touches;
+            },
+            set: function (value) {
+                if (Array.isArray(value) && this._event.touches) {
+                    this._event.touches.splice.apply(this._event.touches, [0, Infinity].concat(value));
+                } else if (Array.isArray(value)) {
+                    this._event.touches = value;
+                }
             }
```
marchant: if (Array.isArray(value) could be done twice, maybe worth refactoring to avoid, same in changedTouches

tejaede: @marchant resolved in both setters

tejaede:
@marchant @johnnykahalawai 

Regarding context, my understanding is that Leaflet sets `event.touches` and `event.changedTouches` and fails if those setters are absent. Jonathan may know more specifics.

johnnykahalawai:
@marchant ,

Without the change, this code: https://github.com/Leaflet/Leaflet/blob/b2dd61d02fea3c5cb8c155737afe8a53ca25d400/src/dom/DomEvent.Pointer.js#L104

Will throw an error:

Cannot set property touches of #<constructor> which has only a getter

marchant:
Thanks for the context, makes sense!

marchant:
> I suppose there is nothing in Montage code that will use the setter? So perhaps we should remove the array check and splice apply altogether?
> 
> I do not see anything in the leaflet code that expects this.
Right, good point, this is new, so no backward compatibility worries here in montage

tejaede:
Makes sense to me. I removed the conditionals from the setters.

PR #2004, opened 2019-01-12
tej/logout

tejaede:
This PR fixes a logical hole in `AuthorizationManager.clearAuthorizationForService()` in which the `AuthorizationManager` does not propagate the clear throughout the service tree. `AuthorizationManager` must (but does not) remove the cleared authorization object from every `DataService` that uses it. This PR includes code that iterates over the affected services and removes their authorization property.

For context, note the following about `aDataService.authorization`: 
The property.....
1. is the authorization object returned from the `AuthorizationManager` when a data service is first authorized. 
2. is used to check whether the service is authorized on fetches.
https://github.com/montagejs/montage/blob/master/data/service/data-service.js#L1803
3. provides the information necessary for the service to connect to its remote counterpart. The content and structure of the object vary based on the implementation, but the most common use case is Authorization holding a username/password pair or an Authorization HTTP header.

___
The PR also includes unrelated changes that were minor enough so as to not warrant their own branch. 
1. Rename `DataService. authorizationManagerWillAuthorizeWithService` to `DataService. authorizationManagerWillAuthorizeWithProvider` to match the `AuthorizationManager` convention. (`AuthorizationManager` will still look for the legacy version in the interest of backwards compatibility)
2. Add spec for `AuthorizationManager.hasPendingServices`
3. Remove a redundant call to `HTMLFragment._sanitizeNode()`


marchant:
Looks good!

PR #2005, opened 2019-01-22
Remove overlays, are identical for node & browser

cdebost:
Related to montagejs/mr#141. Overlay seems to be serving no purpose in this repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,44 +7,13 @@
     "type": "git",
     "url": "https://github.com/montagejs/montage.git"
   },
-  "main": "montage",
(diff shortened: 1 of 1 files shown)


hthetiot:
+1

hthetiot:
Does #1990 use overlay ?

cdebost:
@hthetiot Different kind of overlay. That PR is for the UI overlay component. In this PR, `overlay` refers to the extra package.json property that we use to resolve dependencies differently on browser vs node.

PR #2007, opened 2019-02-04
Allow undefined arguments to be passed to Object.assign

tejaede:
`Object.assign()` should not throw if given undefined or null arguments, per MDN. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign (last line of description section as of this writing)

his PR updates the object polyfill accordingly.

### DIFF
--- a/core/shim/object.js
+++ b/core/shim/object.js
@@ -81,9 +81,11 @@ if (!Object.assign) {
             }
             for (s = 1, ss = arguments.length; s < ss; ++s) {
                 source = arguments[s];
-                keys = source ? Object.keys(source) : [];
-                for (k = 0, kk = keys.length; k < kk; ++k) {
-                    target[keys[k]] = source[keys[k]];
+                if (source) {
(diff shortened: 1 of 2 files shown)


core/shim/object.js
```diff
@@ -81,7 +81,7 @@ if (!Object.assign) {
             }
             for (s = 1, ss = arguments.length; s < ss; ++s) {
                 source = arguments[s];
-                keys = Object.keys(source);
+                keys = source ? Object.keys(source) : [];
```
marchant: Better to avoid creating an empty array in a loop. Better to use a shared instance, or avoid creating an empty array altogether.

tejaede: Good catch! Fixed.

https://github.com/montagejs/montage/pull/2007/commits/a078d127e1e57bb29c47e3838d086677919a8ea9

marchant: 👍🏻

PR #2009, opened 2019-02-12
Remove creation of extra promises from reviver

cdebost:
Currently the reviver takes care of ensuring all of its APIs return promises, even when a value can be revived synchronously. The reviver is really only used by the montage deserializer, and in the future we want to be able to add a sync option to the deserializer to delay the instantiation phase of MJSON compilation. This PR moves the responsibility of promisifying up to the deserializer, so that adding a sync option will be simpler. Along the way, we also avoid creating extra promises where we don't need to.

### DIFF
--- a/core/serialization/deserializer/montage-interpreter.js
+++ b/core/serialization/deserializer/montage-interpreter.js
@@ -130,8 +130,6 @@ var MontageContext = Montage.specialize({
                 objects = this._objects,
                 object;
 
-
-
             if (label in objects) {
                 return objects[label];
             } else if (label in serialization) {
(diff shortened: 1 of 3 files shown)


marchant:
Looks good, but where are the matching changes to promisify in the deserializer, did I miss something? I think I saw something similar in [Add sync option to deserializer #2011](url
https://github.com/montagejs/montage/pull/2011)

cdebost:
@marchant As described in #2011, this PR is a dependency of #2011 and #2011 was meant to be merged on top of this. I thought it would be easier to review the non-deserializer related changes first and then work on #2011 after. But at this point the commits in this PR have already been reviewed and updated in #2011 so we may as well close this.

PR #2010, opened 2019-02-12
Montage data converters#1

marchant:
To merge after /feature/npm3 
- Adds a new new converter to map embedded types 
- fixes a deserialization bug
- adds resolveObjectForTypeRawData method to RawDataService


core/serialization/deserializer/montage-reviver.js
```diff
@@ -502,7 +502,14 @@ var MontageReviver = exports.MontageReviver = Montage.specialize(/** @lends Mont
                 object = value && "prototype" in value ?
                     Object.create(module.montageObject) : module.montageObject;
                 context.setObjectLabel(object, label);
-                return this.instantiateMJSONObject(value, object, objectName, context, label);
+
+                if(label === "root") {
+                    return this.instantiateMontageObject(value, object, objectName, context, label);
+                }
+                else {
+                    return this.instantiateMJSONObject(value, object, objectName, context, label);
```
cdebost: An object doesn't receive `deserializedSelf` if it's the root? Isn't that an issue that should be fixed in `instantiateMJSONObject`?

data/service/data-service.js
```diff
@@ -272,16 +272,21 @@ exports.DataService = Montage.specialize(/** @lends DataService.prototype */ {
     _addChildService: {
         value: function (child, types) {
             var children, type, i, n, nIfEmpty = 1;
-            types = types || child.model && child.model.objectDescriptors || child.types;
+            types = types || child.model && child.model.objectDescriptors || child.types,
+            isNotParentService = (child._parentService !== this);
```
cdebost: Positive variables are generally clearer, i.e. `isParentService` and `!isParentService`.

data/converter/expression-data-mapping-converter.js
```diff
@@ -182,7 +184,10 @@ exports.ExpressionDataMappingConverter = Converter.specialize( /** @lends Expres
 
     compiledRevertSyntax: {
         get: function () {
-            return this._compiledRevertSyntax || (this._compiledRevertSyntax = compile(this.revertSyntax));
+
+            return this._compiledRevertSyntax || (this._compiledRevertSyntax === undefined
+                ? this._compiledRevertSyntax = this.revertSyntax ? compile(this.revertSyntax) : null
+                : null);
```
cdebost: These ternaries are getting too complicated, consider using conditional flow instead.

data/service/raw-data-service.js
```diff
@@ -435,6 +435,12 @@ exports.RawDataService = DataService.specialize(/** @lends RawDataService.protot
             }
             this._addMapDataPromiseForStream(result, stream);
 
+
+            //TO-DO: #warning
```
cdebost: I think we use `TODO` everywhere else. Many editors have automatic support for finding TODO comments in code.

data/converter/raw-embedded-value-to-object-converter.js
```diff
@@ -89,7 +86,7 @@ exports.RawEmbeddedValueToObjectConverter = ExpressionDataMappingConverter.speci
                 });
             }
             else  {
-                result = dataObject;
+                result = Promise.resolve(undefined);
```
cdebost: `Promise.resolve()` is equivalent.

core/serialization/deserializer/montage-reviver.js
marchant: Will look more into it

marchant:
Agreed, how about parent: RawValueToObjectConverter 
then RawEmbeddedValueToObjectConverter and 1. RawForeignValueToObjectConverter ?

marchant:
- refactor/add convertDataQuery and revertDataQuery to adds more transparency and allow upper layers to expose that query.

core/serialization/deserializer/montage-reviver.js
marchant: This has been taken care of

data/converter/expression-data-mapping-converter.js
marchant: I'd rather not, for perf reasons, mappings should be faster

data/service/raw-data-service.js
marchant: taken care of

data/converter/raw-embedded-value-to-object-converter.js
marchant: taken care of

data/service/data-service.js
```diff
@@ -1324,6 +1398,7 @@ exports.DataService = Montage.specialize(/** @lends DataService.prototype */ {
 
 
             if (mapping) {
+                //Why aren't we passing this.snapshotForObject(object) instead of copying everying in a new empty object?
                 Object.assign(data, this.snapshotForObject(object));
```
tejaede: The criteria source was a first attempt to support derived properties. It's a bucket in which we can put data from the cooked object in order to fetch other cooked properties. The cooked data placed in the criteria source does not belong in the snapshot. 

For example, take this model:
- A `Foo` model includes a `bars` property and a `baz` property. 
- `Foo.baz` is derived from the value of `Foo.bars`.

When the application triggers a fetch for `Foo.baz`, the value of `Foo.bars` needs to be available to build the criteria for the fetch. However, the value of `Foo.bars` is cooked and does not belong in the snapshot. Therefore, we create a copy of the snapshot called the "criteria source" into which we can put `Foo.bars`. 

All of this said, I don't know this is the right way to solve the problem. The issue at the moment is that this functionality is being used so we cannot remove it without an alternative.

data/service/data-service.js
```diff
@@ -1332,6 +1407,7 @@ exports.DataService = Montage.specialize(/** @lends DataService.prototype */ {
                         return mapping.mapRawDataToObjectProperty(data, object, propertyName);
                     });
                 } else {
+                    //This was already done a few lines up. Why are we re-doing this?
```
tejaede: I agree, this seems redundant.

data/service/data-service.js
marchant: Added as comment in code as this is important to capture.

data/service/data-service.js
marchant: ok, filed https://github.com/montagejs/montage/issues/2013

marchant:
This merged the latest in #master to get sync capabilities in the deserializer, and it needs https://github.com/montagejs/mr/pull/146 to work, which is itself additional to the npm3 work, so both PR go hand in hand

marchant:
montage tests pass, also tested with a montage data improved version of shop-demo (which wasn't working without these changes as it's the first use of a main.datareel folder that contains all data related assets) as well as Popcorn master. Please review again.

marchant:
Some test fails as it needs https://github.com/montagejs/mr/pull/146, I didn't change package.json, so it still points  to "18.0.0-rc1" where https://github.com/montagejs/mr/pull/146 needs to be merged on

core/meta/module-object-descriptor.js
```diff
@@ -60,6 +60,9 @@ var ModuleObjectDescriptor = exports.ModuleObjectDescriptor = ObjectDescriptor.s
             this.super(serializer);
             this._setPropertyWithDefaults(serializer, "module", this.module);
             this._setPropertyWithDefaults(serializer, "exportName", this.exportName);
+            if(this.object) {
+                this._setPropertyWithDefaults(serializer, "object", this.object);
```
tejaede: @marchant Is there a spec for serializing/deserializing `ModuleObjectDescriptor#object`?

tejaede:
An error with `DataService` deserialization bubbled up some inconsistencies. 

1. `DataService._childServices` is designed to be a `Set`, per the documentation. However, `DataService.deserializeSelf` set the value to an array. 

2. `DataService.deserializedFromSerialization` called `DataService.addChildServices` with `DataService._childServices`. As mentioned above, `DataService._childServices` is an `Array` if the serialization includes child services, but is a `Set` if the value is allowed to lazy load. 

I am unable to write to marchant/montagejs so I pushed fixes to my own repo for these issues for your review: https://github.com/tejaede/montage/commit/982991277699a6f9280ee27995647cdab3686b7f

The fixes are: 
1. Ensure `DataService.childServices` is always a set
2. Use iterator in `DataService.addChildServices` so the argument can be either a `Set` or an `Array`

cc: @marchant

marchant:
There's  a problem running karma on travis, we need to fix this independently of this PR, merging.

PR #2011, opened 2019-02-12
Add sync option to deserializer

cdebost:
Depends on #2009

### DIFF
--- a/core/serialization/deserializer/montage-deserializer.js
+++ b/core/serialization/deserializer/montage-deserializer.js
@@ -25,7 +25,7 @@ var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
     },
 
     init: {
-        value: function (serialization, _require, objectRequires, locationId) {
+        value: function (serialization, _require, objectRequires, locationId, sync) {
             if (typeof serialization === "string") {
                 this._serializationString = serialization;
             } else {
@@ -35,8 +35,9 @@ var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
             this._locationId = locationId ? locationId.indexOf(_require.location) === 0 ? locationId : _require.location + locationId : locationId;
 
             this._reviver = new MontageReviver().init(
-                _require, objectRequires, this.constructor
+                _require, objectRequires, this.constructor, sync
             );
+            this._sync = sync;
 
             return this;
         }
@@ -48,33 +49,49 @@ var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
      * link against the serialization.
      * @param {Element} element The root element to resolve element references
      * against.
-     * @return {Promise}
+     * @return {Promise|object} Deserialized objects if the deserializer was
+     * initialized with sync set to true, or a Promise for the deserialized
+     * objects otherwise.
      */
     deserialize: {
         value: function (instances, element) {
-            var context = this._locationId && MontageDeserializer.moduleContexts.get(this._locationId);
+            var context = this._locationId && MontageDeserializer.moduleContexts.get(this._locationId),
+                circularError;
             if (context) {
                 if (context._objects.root) {
-                    return Promise.resolve(context._objects);
+                    return this._sync ? context._objects : Promise.resolve(context._objects);
                 } else {
-                    return Promise.reject(new Error(
+                    circularError = new Error(
                         "Unable to deserialize because a circular dependency was detected. " +
                         "Module \"" + this._locationId + "\" has already been loaded but " +
                         "its root could not be resolved."
-                    ));
+                    );
+                    if (this._sync) {
+                        throw circularError;
+                    } else {
+                        return Promise.reject(circularError);
+                    }
                 }
             }
 
             try {
                 var serialization = JSON.parse(this._serializationString);
                 context = new MontageContext()
-                    .init(serialization, this._reviver, instances, element, this._require);
+                    .init(serialization, this._reviver, instances, element, this._require, this._sync);
                 if (this._locationId) {
                     MontageDeserializer.moduleContexts.set(this._locationId, context);
                 }
-                return context.getObjects();
+                try {
+                    return context.getObjects();
+                } catch (ex) {
+                    return Promise.reject(ex);
+                }
             } catch (ex) {
-                return this._formatSerializationSyntaxError(this._serializationString);
+                if (this._sync) {
+                    throw ex;
+                } else {
+                    return this._formatSerializationSyntaxError(this._serializationString);
+                }
             }
         }
     },
--- a/core/serialization/deserializer/montage-deserializer.js
+++ b/core/serialization/deserializer/montage-deserializer.js
@@ -25,7 +25,7 @@ var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
     },
 
     init: {
-        value: function (serialization, _require, objectRequires, locationId, sync) {
+        value: function (serialization, _require, objectRequires, locationId, isSync) {
             if (typeof serialization === "string") {
                 this._serializationString = serialization;
             } else {
@@ -35,9 +35,9 @@ var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
             this._locationId = locationId ? locationId.indexOf(_require.location) === 0 ? locationId : _require.location + locationId : locationId;
 
             this._reviver = new MontageReviver().init(
-                _require, objectRequires, this.constructor, sync
+                _require, objectRequires, this.constructor, isSync
             );
-            this._sync = sync;
+            this._isSync = isSync;
 
             return this;
         }
@@ -59,14 +59,14 @@ var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
                 circularError;
             if (context) {
                 if (context._objects.root) {
-                    return this._sync ? context._objects : Promise.resolve(context._objects);
+                    return this._isSync ? context._objects : Promise.resolve(context._objects);
                 } else {
                     circularError = new Error(
                         "Unable to deserialize because a circular dependency was detected. " +
                         "Module \"" + this._locationId + "\" has already been loaded but " +
                         "its root could not be resolved."
                     );
-                    if (this._sync) {
+                    if (this._isSync) {
                         throw circularError;
                     } else {
                         return Promise.reject(circularError);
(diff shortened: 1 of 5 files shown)


core/serialization/deserializer/montage-reviver.js
```diff
@@ -341,7 +337,7 @@ var MontageReviver = exports.MontageReviver = Montage.specialize(/** @lends Mont
             // should be rejected as an error.
             error = this._checkLabel(label, isAlias);
             if (error) {
-                return Promise.reject(error);
+                throw error;
```
marchant: shouldn't there be test if not sync then Promise.reject(error) as before?

core/serialization/deserializer/montage-reviver.js
```diff
@@ -419,7 +415,7 @@ var MontageReviver = exports.MontageReviver = Montage.specialize(/** @lends Mont
                 }
                 return element;
             } else {
-                return Promise.reject(new Error("Element with id '" + elementId + "' was not found."));
+                throw new Error("Element with id '" + elementId + "' was not found.");
```
marchant: shouldn't there be test if not sync then Promise.reject(error) as before?

core/serialization/deserializer/montage-reviver.js
```diff
@@ -469,14 +465,20 @@ var MontageReviver = exports.MontageReviver = Montage.specialize(/** @lends Mont
                 )) {
                 // We have a circular reference. If we wanted to forbid circular
                 // references this is where we would throw an error.
-                return Promise.resolve(this._deserializerConstructor.moduleContexts.get(location)._objects.root);
+                return this._deserializerConstructor.moduleContexts.get(location)._objects.root;
```
marchant: shouldn't that return a promise as before when this._sync is false?

core/serialization/deserializer/montage-deserializer.js
```diff
@@ -35,8 +35,9 @@ var MontageDeserializer = exports.MontageDeserializer = Montage.specialize({
             this._locationId = locationId ? locationId.indexOf(_require.location) === 0 ? locationId : _require.location + locationId : locationId;
 
             this._reviver = new MontageReviver().init(
-                _require, objectRequires, this.constructor
+                _require, objectRequires, this.constructor, sync
             );
+            this._sync = sync;
```
marchant: _isSync would be a better name

core/serialization/deserializer/montage-interpreter.js
```diff
@@ -170,8 +172,10 @@ var MontageContext = Montage.specialize({
             }
 
             if (promises.length === 0) {
-                return Promise.resolve(this._invokeDidReviveObjects());
+                result = this._invokeDidReviveObjects();
```
marchant: How is that working if the for() above happens to fill promises array while the interpreter has been set to be synchronous?

core/serialization/deserializer/montage-interpreter.js
```diff
@@ -208,12 +212,8 @@ var MontageContext = Montage.specialize({
                 result;
 
             if (typeof reviver.didReviveObjects === "function") {
-                result = reviver.didReviveObjects(this._objects, this);
-                if (Promise.is(result)) {
-                    return result.then(function() {
-                        return self._objects;
-                    });
-                }
+                reviver.didReviveObjects(this._objects, this);
```
marchant: what if the interpreter isn't synchronous?

core/serialization/deserializer/montage-reviver.js
```diff
@@ -469,14 +465,20 @@ var MontageReviver = exports.MontageReviver = Montage.specialize(/** @lends Mont
...
             }
 
             if (Promise.is(module)) {
+                if (this._sync) {
+                    throw new Error(
+                        "Tried to revive montage object with label " + label +
+                        " synchronously but the module was not loaded: " + JSON.stringify(value)
+                    );
+                }
                 return module.then(function (exports) {
```
marchant: How can this still be async if the reviver is sync?

core/serialization/deserializer/montage-reviver.js
```diff
@@ -687,14 +689,9 @@ var MontageReviver = exports.MontageReviver = Montage.specialize(/** @lends Mont
...
         value: function (objects, context) {
-            var self = this;
-
-            return Promise.all([
-                this._deserializeBindings(context),
-                this._deserializeUnits(context)
-            ]).then(function () {
-                self._invokeDeserializedFromSerialization(objects, context);
-            });
+            this._deserializeBindings(context);
```
marchant: Where are the promises in async case?

core/serialization/deserializer/montage-reviver.js
```diff
@@ -731,17 +728,13 @@ var MontageReviver = exports.MontageReviver = Montage.specialize(/** @lends Mont
...
-                        bindingsToDeserializeDesc = bindingsToDeserialize[i];
-                        Bindings.deserializeObjectBindings(
-                            unitDeserializer.initWithContext(context),
-                            bindingsToDeserializeDesc.object,
-                            bindingsToDeserializeDesc.bindings
-                        );
-                    }
-                } catch (ex) {
-                    return Promise.reject(ex);
+                for (var i = 0, length = bindingsToDeserialize.length; i < length; i++) {
```
marchant: how is that backward compatible with default being async?

core/serialization/deserializer/montage-reviver.js
```diff
@@ -753,19 +746,15 @@ var MontageReviver = exports.MontageReviver = Montage.specialize(/** @lends Mont
                 unitDeserializer = new UnitDeserializer(),
                 unitNames;
 
-            try {
-                for (var i = 0, unitsDesc; (unitsDesc = unitsToDeserialize[i]); i++) {
-                    unitNames = unitsDesc.unitNames;
+            for (var i = 0, unitsDesc; (unitsDesc = unitsToDeserialize[i]); i++) {
```
marchant: how is that backward compatible with default being async?

core/serialization/deserializer/montage-interpreter.js
cdebost: Because `getObject` will throw an error instead of returning a promise if the context has been set to sync but some async work needs to be done.

core/serialization/deserializer/montage-interpreter.js
cdebost: In practice, `didReviveObjects` never actually did any async work but was returning a promise anyway.

core/serialization/deserializer/montage-reviver.js
cdebost: I don't think it's necessary because `reviveMontageObject` is called by `reviveRootObject` which is called by `getObject`. `getObject` already assumes that the "object" it gets back could be a promise or a value and handles it appropriately. I think this is a case of an extra promise that we don't need.

core/serialization/deserializer/montage-deserializer.js
cdebost: ok

core/serialization/deserializer/montage-reviver.js
cdebost: I don't understand the question. The intention here is that if we are sync but the module we are loading is async, we need to throw an error.

core/serialization/deserializer/montage-reviver.js
cdebost: In practice, both _deserializeBindings and _deserializeUnits don't do any async work. So instead of creating extra promises for no reason, we can make this method synchronous.

core/serialization/deserializer/montage-reviver.js
cdebost: There is no async work being done here, we don't need to wrap the error in a promise.

core/serialization/deserializer/montage-reviver.js
cdebost: There is no async work being done here, we don't need to wrap the error in a promise.

core/serialization/deserializer/montage-deserializer.js
cdebost: ok

core/serialization/deserializer/montage-interpreter.js
cdebost: Because `getObject` will throw an error instead of returning a promise if the context has been set to sync but some async work needs to be done.

core/serialization/deserializer/montage-interpreter.js
cdebost: In practice, `didReviveObjects` never actually did any async work but was returning a promise anyway.

core/serialization/deserializer/montage-reviver.js
cdebost: I don't think it's necessary because `reviveMontageObject` is called by `reviveRootObject` which is called by `getObject`. `getObject` already assumes that the "object" it gets back could be a promise or a value and handles it appropriately. I think this is a case of an extra promise that we don't need.

core/serialization/deserializer/montage-reviver.js
cdebost: I don't understand the question. The intention here is that if we are sync but the module we are loading is async, we need to throw an error.

core/serialization/deserializer/montage-reviver.js
cdebost: In practice, both _deserializeBindings and _deserializeUnits don't do any async work. So instead of creating extra promises for no reason, we can make this method synchronous.

core/serialization/deserializer/montage-reviver.js
cdebost: There is no async work being done here, we don't need to wrap the error in a promise.

core/serialization/deserializer/montage-reviver.js
cdebost: There is no async work being done here, we don't need to wrap the error in a promise.

core/serialization/deserializer/montage-reviver.js
cdebost: It's not strictly necessary because the montage-deserializer does a try/catch over context API calls to catch synchronous errors as well as rejected promises. Since nothing but the deserializer uses context APIs, why create an extra promise when we could just throw a regular error?

core/serialization/deserializer/montage-reviver.js
cdebost: It's not strictly necessary because the montage-deserializer does a try/catch over context API calls to catch synchronous errors as well as rejected promises. Since nothing but the deserializer uses context APIs, why create an extra promise when we could just throw a regular error?

core/serialization/deserializer/montage-reviver.js
marchant: Doesn't context._require.async(locationDesc.moduleId); returns a promise anyway? Shouldn't require() be used if sync instead?

cdebost: Yes. Looks like the condition on 471 means the module will be loaded async if it is an .mjson but has not been compiled into a `montageObject` yet. If we are sync and we fall into that condition, it means our caller did not take care of loading all of our dependencies beforehand, and we should throw.

cdebost: @marchant Ok with this?

marchant: yep, merged it!

PR #2012, opened 2019-02-28
Add burmese language to messageformat

tejaede:
Rule logic generated based on the burmese rules defined on unicode.org:
screen shot 2019-02-28 at 9 20 29 am
Link: https://www.unicode.org/cldr/charts/34/supplemental/language_plural_rules.html


marchant:
LGTM, nice addition

